PR #7, opened 2026-04-20
fix: scope token usage data to own key for API key users

XiaoYingGee:
## Summary
- API key (non-admin) users viewing the usage dashboard could see records for **all** keys
- Server-side: `token-usage.ts` now forces `key_id` to the authenticated user's own key when `isAdmin` is false
- Client-side: `client.tsx` passes `key_id` in the fetch URL for non-admin sessions as defense-in-depth

## Test plan
- [ ] Login with admin key → usage page shows all keys
- [ ] Login with API key → usage page shows only that key's data
- [ ] `curl /api/token-usage` with API key auth → response scoped to that key only

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,40 @@
+# Build context exclusions for the Docker image.
+# Keep image small + don't leak secrets/runtime data.
+
+.git
+.gitignore
+.github
+.claude
+.wrangler
+.vscode
+.idea
+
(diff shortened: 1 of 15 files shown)


XiaoYingGee:
Closing: PR contained unrelated commits from fork. Will resubmit with only the scoping fix.

PR #8, opened 2026-04-20
fix: scope token usage to authenticated API key

XiaoYingGee:
## Problem

When a non-admin user logs into the dashboard with an API key, the Usage tab shows token consumption for **all** keys in the system. This leaks usage information across tenants — a user with key "Desktop" can see exactly how much "WorkStation", "Hermes", etc. are consuming.

The original design comment in `token-usage.ts` stated this was intentional, but in practice it breaks the multi-tenant isolation that the API key system provides. Admin users should see everything; API key users should only see their own data.

## Changes

**Server-side (`src/routes/token-usage.ts`):**
- Non-admin requests now ignore the `key_id` query parameter and are forced to their own `apiKeyId` (set by auth middleware)
- Admin requests retain the existing behavior (optional `key_id` filter, or all keys if omitted)

**Client-side (`src/ui/dashboard/client.tsx`):**
- Non-admin sessions pass `key_id` in the fetch URL as defense-in-depth
- Reads `login_key_id` from localStorage (already stored by the login flow but previously unused)

## Test plan
- [ ] Login with admin key → usage page shows all keys
- [ ] Login with API key → usage page shows only that key's data
- [ ] `curl /api/token-usage` with API key in `x-api-key` header → response scoped to that key only
- [ ] `curl /api/token-usage?key_id=other` with API key → `key_id` param ignored, still returns own data only

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/routes/token-usage.ts
+++ b/src/routes/token-usage.ts
@@ -1,15 +1,16 @@
 // GET /api/token-usage — query per-key token usage records
 //
-// IMPORTANT DESIGN DECISION: Usage data is intentionally readable by ALL authenticated
-// users (both admin and API key users), without scoping. Any authenticated user can view
-// usage records for all keys. API keys themselves are only readable by their owner.
+// Admin users can view all keys or filter by key_id.
+// API key users are scoped to their own key only.
 
 import type { Context } from "hono";
 import { queryUsage } from "../lib/usage-tracker.ts";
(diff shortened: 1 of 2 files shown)


XiaoYingGee:
Closing — upstream's design decision to expose all usage data to any authenticated user is intentional. Keeping this scoping fix in our fork only.

PR #9, opened 2026-04-22
fix: strip Claude Code billing attribution header to restore prompt cache

oc-pegasus:
## Problem

Claude Code's native binary injects a billing attribution header into the system prompt:

```
x-anthropic-billing-header: cc_version=2.1.114; cc_entrypoint=cli; cch=<hash>
```

The `cch=` hash is computed from the request body and changes with **every request**. Since the Copilot API uses prefix-based prompt caching, the changing hash causes the system prompt prefix to differ on every turn, completely invalidating prompt cache.

**Impact**: `cache_read_input_tokens` is always 0 for all Claude Code sessions, while `cache_creation_input_tokens` shows tokens being cached but never reused.

## Solution

- Upgrade `stripReservedKeywords` to remove the entire billing header line (not just the keyword)
- Strip orphaned `cch=<hex>` hashes from message content
- Filter empty text blocks after stripping to prevent Anthropic API 400 errors ("text content blocks must be non-empty")
- Remove system entirely when all blocks are stripped (billing-only system prompts)

## Verification

- Confirmed cache hit restoration: after fix, `cache_read_input_tokens` goes from 0 to matching `cache_creation_input_tokens` on subsequent turns
- Added 4 unit tests covering: string system, array system with mixed blocks, all-billing system blocks, and message content stripping

### DIFF
--- a/src/routes/messages.ts
+++ b/src/routes/messages.ts
@@ -113,17 +113,18 @@ const BILLING_HEADER_LINE_RE =
 const CCH_HASH_RE = /cch=[0-9a-f]{5,};?/gi;
 
 function stripBillingAttribution(text: string): string {
-  let result = text
+  return text
     .replace(BILLING_HEADER_LINE_RE, "")
     .replace(CCH_HASH_RE, "")
     .trim();
-  // Never return empty — Anthropic API rejects empty text content blocks
-  return result || " ";
 }
 
 function stripReservedKeywords(payload: AnthropicMessagesPayload): void {
   if (typeof payload.system === "string") {
     payload.system = stripBillingAttribution(payload.system);
+    if (!payload.system) {
+      delete (payload as Record<string, unknown>).system;
+    }
   } else if (Array.isArray(payload.system)) {
     for (const block of payload.system) {
       block.text = stripBillingAttribution(block.text);
--- a/src/routes/messages.ts
+++ b/src/routes/messages.ts
@@ -93,26 +93,61 @@ function isContextWindowError(text: string): boolean {
     text.includes("context_length_exceeded");
 }
 
-/** Copilot rejects requests containing this string in system prompts */
-const RESERVED_KEYWORD = "x-anthropic-billing-header";
+/**
+ * Strip the entire billing attribution line from Claude Code.
+ *
+ * Claude Code's native binary injects a billing header into the system prompt:
+ *   x-anthropic-billing-header: cc_version=2.1.114; cc_entrypoint=cli; cch=<hash>
+ *
+ * The `cch=` hash changes with EVERY request (it's a hash of the request body),
+ * which means the system prompt content changes every turn, completely invalidating
+ * the prompt cache. The previous code only stripped the keyword itself, leaving
+ * the varying hash in the text.
+ *
+ * This regex removes the entire line (keyword + all parameters including the hash).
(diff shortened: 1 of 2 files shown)


src/routes/messages.ts
```diff
@@ -93,26 +93,61 @@ function isContextWindowError(text: string): boolean {
...
+ * This regex removes the entire line (keyword + all parameters including the hash).
+ * We also strip any orphaned `cch=<hex>` patterns that the native binary's memmem
+ * might have injected into message content instead of the billing header.
+ */
+const BILLING_HEADER_LINE_RE =
+  /x-anthropic-billing-header[^\n]*/g;
+const CCH_HASH_RE = /cch=[0-9a-f]{5,};?/gi;
+
+function stripBillingAttribution(text: string): string {
+  let result = text
```
Menci: Why `let` for an immutable variable?

src/routes/messages.ts
```diff
@@ -93,26 +93,61 @@ function isContextWindowError(text: string): boolean {
...
   if (typeof payload.system === "string") {
-    payload.system = payload.system.replaceAll(RESERVED_KEYWORD, "");
+    payload.system = stripBillingAttribution(payload.system);
   } else if (Array.isArray(payload.system)) {
     for (const block of payload.system) {
-      block.text = block.text.replaceAll(RESERVED_KEYWORD, "");
+      block.text = stripBillingAttribution(block.text);
+    }
+    // Remove blocks that are now effectively empty (only whitespace)
+    payload.system = payload.system.filter((b) => b.text.trim().length > 0);
```
Menci, line 133: ???

`result || " "`

and then

`trim()`

Menci (changes requested):

src/routes/messages.ts
oc-pegasus: Good catch — fixed. Refactored to a direct return chain, no variable needed at all now.

src/routes/messages.ts
oc-pegasus: You're right, this was contradictory — returning ` ` as fallback but then filtering by `trim().length > 0` downstream. Removed the fallback entirely; empty results are now handled in `stripReservedKeywords` which filters out empty blocks and deletes system when all blocks are gone. Also added handling for string-type system that becomes empty after stripping.

oc-pegasus:
Thanks for the detailed feedback. You're absolutely right — the original code had inconsistent behavior between the two branches:

- **String system**: `stripBillingAttribution` returned `" "` (space fallback) → space stayed as the system prompt
- **Array system**: downstream `filter(b => b.text.trim().length > 0)` removed the space block → `delete system`

Same empty input, different outcomes. That's a real bug, not just a style issue.

**The correct behavior is `undefined`** — when system content is fully stripped, it should be deleted, not replaced with a meaningless space. A space in the system prompt wastes tokens and has no semantic value.

This is fixed in the latest commit (`189c185`):
- `stripBillingAttribution` returns the trimmed result directly (can be empty string, no fallback)
- String system: empty → `delete payload.system`
- Array system: filter empty blocks → all gone → `delete payload.system`

Both paths now consistently produce `undefined` for empty system.

Menci:
Merged manually after resolving conflicts and adjusting to only strip billing attribution from system prompts (not messages). Squashed into a single commit. Thanks!

PR #11, opened 2026-04-30
fix(data-plane): include usageModel metadata in embeddings serve

XiaoYingGee:
## Summary

`POST /v1/embeddings` returns HTTP 500 for every authenticated request after d8dd086 (`fix(data-plane): account usage under resolved model`). All embeddings traffic going through the gateway is broken; chat-class endpoints are unaffected.

```
Error: Usage response has token usage but no model
    at requireUsageModel (src/middleware/usage.ts:155:9)
```

## Root cause

d8dd086 added `requireUsageModel()` in the usage middleware and updated the chat-completions, messages, and responses serves to write `usageModel` metadata via `withUsageResponseMetadata()`. `src/data-plane/embeddings/serve.ts` was not updated.

GitHub Copilot's `/embeddings` upstream response carries `usage` but no `model` field, so when the usage middleware extracts tokens from the JSON body it has no model signal in either the metadata header or the response payload, and `requireUsageModel()` throws.

## Reproduction

```bash
curl -X POST https://<gateway>/v1/embeddings \
  -H "Authorization: Bearer <user-api-key>" \
  -H "Content-Type: application/json" \
  -d '{"model":"text-embedding-3-small","input":["test"]}'
# HTTP 500 — Internal Server Error
```

## Fix

Mirror the chat-class pattern in `embeddings/serve.ts`: after proxying the upstream response, wrap it with `withUsageResponseMetadata({ usageModel })` using the model from the parsed request body. Embeddings have no alias-resolution layer, so the requested model **is** the resolved model.

```ts
const proxied = proxyJsonResponse(resp);
return model ? withUsageResponseMetadata(proxied, { usageModel: model }) : proxied;
```

If the request body has no parseable `model` field we leave the proxied response untouched and let `requireUsageModel()` continue to surface the malformed-input case (existing behavior).

## Tests

Added a regression test in `src/middleware/usage_test.ts` that mocks an upstream `/embeddings` response **without** a `model` field (matching real Copilot behavior) and asserts the request succeeds with usage recorded under the requested model.

- The new test fails on the parent of this commit (HTTP 500).
- All 522 existing tests still pass.

```
ok | 522 passed | 0 failed
```

## Scope

Single-file fix + single-test addition. No other behavior changes.

### DIFF
--- a/src/data-plane/embeddings/serve.ts
+++ b/src/data-plane/embeddings/serve.ts
@@ -8,11 +8,12 @@ import {
   getErrorMessage,
   proxyJsonResponse,
 } from "../shared/http/proxy-response.ts";
+import { withUsageResponseMetadata } from "../../middleware/usage-response-metadata.ts";
 
 export const embeddings = async (c: Context) => {
   try {
     const body = await c.req.text();
-    let model = "unknown";
+    let model: string | null = null;
     try {
       const parsed = JSON.parse(body) as { model?: unknown };
       if (typeof parsed.model === "string") model = parsed.model;
@@ -20,7 +21,7 @@ export const embeddings = async (c: Context) => {
       // Let upstream preserve the request-shape error; fallback simply has no model signal.
     }
 
-    const resp = await withAccountFallback(model, ({ account }) =>
+    const resp = await withAccountFallback(model ?? "unknown", ({ account }) =>
       copilotFetch(
         "/embeddings",
         { method: "POST", body },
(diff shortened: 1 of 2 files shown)


XiaoYingGee:
Confirming this fix works in production. Cherry-picked into our fork (XiaoYingGee/copilot-gateway) and deployed to a self-hosted Deno gateway behind Cloudflare tunnel.

**Production HEAD:** `d8dd086` (regression, broken) → `7de92d0` (with this fix)

**Verification:**
- `POST /v1/embeddings` (model=`text-embedding-3-small`, input=array) → HTTP 200, ~19.4KB response with real embedding vectors
- Verified on both `localhost:8000` (direct) and the public domain (via Cloudflare tunnel) — both 200
- Real production traffic after restart: 8 consecutive embeddings calls across multiple API keys (including an internal PoC key) all returned 200, with **zero `requireUsageModel` errors** in systemd journal

Fix is effective. Recommend merging. Thanks @lutsky-igor 🙏

XiaoYingGee:
Update: upstream already shipped an equivalent fix in commit f00abdb (`fix(data-plane): support scalar embeddings input`) which addresses the `requireUsageModel` 500 error via `withUsageResponseMetadata` + also fixes a scalar input bug. Production has been switched to upstream main HEAD (0b7e687) and verified OK (10+ /v1/embeddings calls returning 200, no `requireUsageModel` errors in journal). Suggest closing this PR. cc @Menci

XiaoYingGee:
Hi @Menci — thanks for landing the equivalent fix in `f00abdb fix(data-plane): support scalar embeddings input` (main, 2026-05-02), which also catches an additional scalar-input edge case we hadn't covered here.

Our production fork has rebased onto upstream HEAD `0b7e687` and verified end-to-end. Closing from our side, thanks for the equivalent fix! 🙏

PR #15, opened 2026-05-23
fix(scripts): target remote D1 in db:migrate

Elity:
The production deploy workflow in AGENTS.md is documented as:

    pnpm run db:migrate && pnpm run deploy

so db:migrate must target the same environment as the deploy: the remote D1 instance. Without --remote, wrangler defaults to a local sqlite, which silently no-ops against production. This produced a dangerous middle state during deploy: the new Worker code was published while the remote D1 still ran the pre-migration schema.

Local development can still migrate the local D1 directly with:

    pnpm exec wrangler d1 migrations apply copilot-db

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "wrangler dev",
     "deploy": "wrangler deploy",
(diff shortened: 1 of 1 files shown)


Menci:
This has been fixed in 89f06d245cf4ff60fd4734dd1afe6ee76955b496

PR #46, opened 2026-06-05
refactor(data-plane): per-protocol entry redesign

Menci:
## Summary

Replaces the legacy `sources/<P>/traits.ts` + `targets/<P>/emit.ts` machinery with a per-protocol "serve + attempt + adapter" model. Each protocol (responses, messages, chat-completions, gemini) lives in its own directory under `packages/proxy/src/data-plane/llm/<P>/` with a single interceptor list and owns its full lifecycle. `responsesAttempt` is the sole owner of Responses-side adaptation (rewrite + dispatch + wrap-output-storage), entered the same way from native HTTP/WS and from cross-source-via-translation paths.

- Single entry per protocol: HTTP/WS adapter → `<P>Serve.{generate,countTokens,compact}` → `<P>Attempt.<endpoint>`. Cross-protocol translation goes through `traverseTranslation` and the existing pure translate pairs.
- One interceptor list per protocol (source/target split removed). The merged list lives as a module-level constant under `packages/proxy/src/data-plane/llm/<P>/interceptors/index.ts` and runs once before the protocol leaves the gateway. Each provider runs its own boundary interceptor chain inside its `call*` method — there is no longer a `ProviderInterceptors` field on `ProviderModelRecord`; the data plane and provider package are decoupled on the interceptor surface.
- Responses' stateful store is constructed by each entry adapter (`createNonResponsesSourceStore` / `createResponsesHttpStore` / `createResponsesWsSession`), so non-Responses sources stop carrying a no-op store. `RequestContext`/`Invocation<TPayload>` are gone — replaced by `GatewayCtx` and slim per-protocol `<P>Invocation` shapes.
- Empirical drop of `withSafetyIdentifierStripped` and the entire `sourceApi` plumbing: a direct probe against Copilot's `/responses` confirmed the field is silently accepted, not rejected. The interceptor was VSCode-fingerprint mimicry, not protocol correctness.
- `callResponsesCompact` now runs the full Copilot Responses boundary chain (`COPILOT_RESPONSES_COMPACT_BOUNDARY`: inline-image compression, `service_tier` / `image_generation` stripping, `store: false` forcing, `x-vision-request` / `x-initiator` header injection). Main ran zero interceptors on the compact path, leaving it inconsistent with the streaming Responses path; the chain restores parity.
- `stripBillingAttribution` moved from the Copilot-only Messages source chain to the gateway-core `messagesInterceptors` list. It strips Claude-Code-client artifacts (`x-anthropic-billing-header: <token>` lines and `cch=<hash>` markers) that the client inlines into the `system` prompt — opaque to every upstream and poisonous to prompt-cache prefix hashes. Now applies to all Messages-protocol upstreams (copilot / custom / azure).

Spec: `docs/superpowers/specs/2026-06-04-protocol-entry-redesign-design.md`. Plan: `docs/superpowers/plans/2026-06-04-protocol-entry-redesign.md`.

## Test plan

- [x] `pnpm run typecheck` — clean across all 14 workspace projects
- [x] `pnpm run lint` — zero errors
- [x] `pnpm run test` — 1666 tests pass
- [x] `pnpm run dev` boot — Worker ready on 8788
- [x] `GET /v1/models` — 200, lists models
- [x] `POST /v1/messages` streaming + non-streaming on `claude-haiku-4-5` — 200, SSE + JSON envelope shapes
- [x] `POST /v1/messages/count_tokens` — 200, `{"input_tokens":9}`
- [x] `POST /v1/chat/completions` streaming + non-streaming — 200
- [x] `POST /v1/responses` streaming + non-streaming on `gpt-5.4` — 200
- [x] `POST /v1/responses/compact` on `gpt-5.4` — 200, `"object":"response.compaction"` envelope with `input_text` + `encrypted_content` chunks, `copilot_usage.token_details` propagated
- [x] `/v1/responses` HTTP `previous_response_id` round-trip — turn 1 stored with `store: true`, turn 2 referencing the prior id correctly replayed context ("azure")
- [x] `/v1/responses` WebSocket session — turn 1 stored, turn 2 with `previous_response_id` returned the prior context ("BANANA-77"); both turns ended with `response.completed` + WS `response.done` envelope carrying `{id, usage}`
- [x] `POST /v1beta/models/gemini-2.5-pro:generateContent` — 200, valid `usageMetadata` + `candidates`
- [x] `POST /v1beta/models/gemini-2.5-pro:streamGenerateContent?alt=sse` — 200 SSE
- [ ] `POST /v1beta/models/gemini-2.5-pro:countTokens` — **upstream-blocked**: Copilot's `/v1beta/models/<any-gemini>:countTokens` returns `400 INVALID_ARGUMENT "Model X does not support countTokens"` for every exposed Gemini model on the test account. Gateway proxies the error verbatim in Gemini's native shape — not a gateway regression
- [x] `safety_identifier` passthrough on `/v1/responses` — 200; gateway no longer strips the field, Copilot accepts it silently (confirmed via direct dispatch + via the proxied path)
- [x] Multi-provider fallback (first candidate non-events ExecuteResult → verify NOT advancing) — seeded a broken Copilot upstream with invalid `githubToken` at lower sortOrder, sent requests through all four protocols (`/v1/messages`, `/v1/responses`, `/v1/chat/completions`, `/v1beta/models/...:generateContent`); each returned the broken candidate's 401 verbatim (Gemini wraps in its RPC envelope) with ~500ms latency consistent with a single upstream round-trip. Spec conforming: gateway did NOT advance to the working second Copilot.

### Side findings observed during probing

- `POST /v1/responses` with a chat-only model (e.g., `gpt-4.1`, which Copilot exposes without `supported_endpoints` and `capabilities.type=chat`) falls through to `translateResponsesToChatCompletions` and returns 200, as designed. Earlier draft of this section incorrectly reported a 400 — re-probed and corrected.
- WS-only `response.done` summary frame emitted after `response.completed`, matching the documented WS contract.


Menci:
Addressed review feedback:

**Real bug (1.1)** — `bd6dc6a` `fix(responses): seed privatePayload + clear staged output via store.beginAttempt in responsesAttempt`. Confirmed `beginAttempt` had zero production callers, breaking cross-turn web-search shim replay. `rewriteResponsesItemsForCandidate` now returns `{ payload, references }`; `responsesAttempt.{generate,compact}` call `store.beginAttempt(rewritten.references)` after rewrite. New e2e test asserts `getPrivatePayload(wireId)` is seeded after the dispatch.

**PR description (2)** — Edited via `gh pr edit`. Summary bullet now reflects the actual current shape: per-protocol interceptors are module-level constants under `packages/proxy/src/data-plane/llm/<P>/interceptors/index.ts`; providers run their own boundary chains inside `call*`; no `ProviderInterceptors` field on `ProviderModelRecord` (dropped in `08820b9`), no `messagesTarget` slot (dropped in `ee2e0d5`).

**Gemini interceptor unit tests (1.2)** — `00d5110`. Restored per-interceptor tests under `gemini/interceptors/`: strip-safety-settings (2), strip-unsupported-part-fields (3), strip-unsupported-tools (3), suppress-thought-parts (5).

**Secondary coverage gaps (5)**:
- `6d6d799` — chat-completions interceptor tests restored: normalize-usage (5), vendor-deepseek (12), vendor-kimi (2), vendor-qwen (3), disable-reasoning-on-forced-tool-choice (4), include-usage-stream-options (3).
- `a8bf860` — telemetry / stream termination invariants restored at `data-plane/llm/shared/respond_test.ts` (14 tests covering `SourceStreamState.failedAfter` classification, `rememberUsage` filtering, `recordPerformance`/`recordUsage` happy paths + no-op invariants). Three legacy cases (EOF-without-terminal, DONE-only-for-chat, SSE-error-as-failure) are now distributed across per-protocol `events/to-result_test.ts` files; coverage preserved but not in a single file.
- `b57c3ed` — runInterceptors compose-order / retry / throw-propagation tests at `packages/interceptor/src/index_test.ts` (4 tests). Required adding a vitest config + workspace project wiring for that package.
- `24936d8` — lint follow-up on the telemetry test file (import-order autofix).

Final: 1666 tests / 0 fail / 0 skip; lint + typecheck clean.

PR #70, opened 2026-06-19
refactor(provider): unify inbound headers as opts.headers (drop 4th/5th args)

Menci:
## Summary

Consolidates the three current inbound-headers conduits on the provider boundary into a single `opts.headers: Headers`. Stacked on #62.

Before: `callMessages` had `headers` (4th arg, Record), `anthropicBeta` (5th arg, array), and `opts.clientRequestHeaders` (Headers|Record). Three things saying "inbound headers". Now unified:

```ts
callMessages(model, body, signal, opts)
// opts.headers: Headers (always present)
```

### Migration

- All `call*` methods on `ModelProvider` lose their 4th `headers` and 5th `anthropicBeta` parameters; provider impls read everything off `opts.headers`.
- Provider boundary-ctx `headers` fields become `Headers`; interceptors call `.set/.get/.delete` on `opts.headers` instead of Record indexing.
- `UpstreamFetchOptions.extraHeaders` is `Headers` too; upstream fetch helpers iterate it directly.
- Copilot's `withAnthropicBetaHeaderFiltered` now `headers.delete('anthropic-beta')` before its policy branches — with the unified bag seeded from the inbound header, leaving the unfiltered raw value would 400 against Copilot's allow-list.
- `mergeAnthropicBetaHeader` is dead — Azure and custom just pass `opts.headers` through.
- Messages attempt encodes inbound `anthropic-beta` (read from `ctx.clientRequestHeaders`) onto `invocation.headers` so it rides the same bag the interceptor chain mutates. The HTTP-layer `parseAnthropicBeta` helper is gone; Copilot rebuilds the typed slice from `opts.headers.get('anthropic-beta')` at provider entry for its variant-selection branch.
- `noopUpstreamCallOptions.headers` is a getter returning a fresh `Headers` per access so tests that mutate the bag do not bleed state.

### Naming

Per review direction: field is `headers`; context infers in/out direction (in `UpstreamCallOptions` inbound, in `ProviderStreamResult.ok:true.headers` outbound — see #64).

## Stack

Stacked on #62 (introduces the `clientRequestHeaders` field this PR renames + unifies). GitHub auto-retargets to `main` when #62 merges.

## Test plan

- [x] typecheck clean
- [x] 2704 vitest tests passing
- [x] eslint clean


Menci:
Folded into #62; the unified opts.headers + dropped 4th/5th args + interceptor updates now ship as a single coherent PR.

PR #78, opened 2026-06-22
fix(copilot): unblock Pro+/Max/EDU subscribers and harden socket reuse

anagentnetbot-ops:
## Summary

Fixes three related issues with the Copilot provider when used with GitHub Copilot Pro+ subscriptions:

### 1. Unrecognized `individual_pro` account type

GitHub now returns `"individual_pro"` (instead of `"individual"`) for Copilot Pro+ subscribers in the `/copilot_billing/selected_teams` endpoint. This caused `detectAccountType()` to throw:

```
Error: Unknown GitHub Copilot plan: individual_pro
```

**Fix:** Add an alias map (`COPILOT_ACCOUNT_TYPE_ALIASES`) that normalizes `individual_pro` → `individual`, and export a `normalizeCopilotAccountType()` helper used by the device-flow auth.

### 2. Incorrect API base URL for individual accounts

The individual endpoint was hardcoded to `https://api.githubcopilot.com`, which is deprecated for Pro+ accounts. The correct endpoint is `https://api.individual.githubcopilot.com`.

**Fix:** Update `COPILOT_BASE_URLS.individual` to the new URL.

### 3. Socket reuse errors (`other side closed`)

Copilot's upstream (`api.individual.githubcopilot.com`) closes keep-alive connections after a single request-response exchange. The default undici connection pool reuses the dead socket, causing:

```
UND_ERR_SOCKET: other side closed
```

Additionally, interceptors may mutate the request body after `content-length` is set, causing `RequestContentLengthMismatchError`.

**Fix:**
- Set `keepAliveTimeout: 1000` and `pipelining: 0` on the global undici dispatcher to force fresh connections.
- Delete the `content-length` header before fetch so undici recomputes it from the actual body.

## Testing

Tested with a Copilot Pro+ subscription — all three issues are resolved and chat completions work correctly through the provider.

### DIFF
--- a/.agents/skills/probing-copilot/SKILL.md
+++ b/.agents/skills/probing-copilot/SKILL.md
@@ -47,7 +47,7 @@ method is GET, not POST — POST returns 404 from this endpoint.
 
 Base URL by account type:
 
-- `individual`  → `https://api.githubcopilot.com`
+- `individual`  → `https://api.individual.githubcopilot.com`
 - `business`    → `https://api.business.githubcopilot.com`
 - `enterprise`  → `https://api.enterprise.githubcopilot.com`
 
--- a/apps/platform-node/entry.ts
+++ b/apps/platform-node/entry.ts
@@ -2,23 +2,11 @@ import { serve, upgradeWebSocket } from '@hono/node-server';
 import { Agent, Pool, setGlobalDispatcher } from 'undici';
 import { WebSocketServer } from 'ws';
 
-// api.{individual,business,enterprise}.githubcopilot.com closes its
-// keep-alive socket immediately after each response. When undici's default
-// pool reuses that socket for the next POST, the race surfaces as either
-// UND_ERR_SOCKET (close lands before the next write) or
-// RequestContentLengthMismatchError (close lands mid-write — see client-h1.js
-// L1340 in undici v6, where bytesWritten < contentLength because the socket
-// was destroyed while the body was still being streamed). Disabling HTTP/1.1
-// connection reuse for these hosts removes the race.
+// api.{individual,business,enterprise}.githubcopilot.com closes its keep-alive
+// socket right after each response; reusing it surfaces as UND_ERR_SOCKET or
+// RequestContentLengthMismatchError. `pipelining: 0` disables keep-alive.
 //
-// `pipelining: 0` is the actual lever per undici's Client docs: "Set to `0`
-// to disable keep-alive connections." The Agent factory restricts the
-// override to the three Copilot data-plane origins so OpenAI, Azure,
-// Gemini, custom OpenAI-compatible, Ollama, GitHub OAuth, and the Codex
-// catalog endpoint all keep undici's default keep-alive behaviour and don't
-// pay an extra TCP+TLS handshake per request.
-//
-// Refs: https://github.com/nodejs/undici/blob/main/docs/docs/api/Client.md
+// Refs: https://github.com/nodejs/undici/blob/v6.21.0/docs/docs/api/Client.md#parameter-clientoptions
 //       https://github.com/Menci/Floway/pull/78#issuecomment-4765475966
 const COPILOT_HOSTS = new Set([
   'api.individual.githubcopilot.com',
@@ -28,7 +16,7 @@ const COPILOT_HOSTS = new Set([
 setGlobalDispatcher(new Agent({
   factory: (origin, opts) => {
     const hostname = typeof origin === 'string' ? new URL(origin).hostname : origin.hostname;
-    return COPILOT_HOSTS.has(hostname) ? new Pool(origin, { ...opts, pipelining: 0 }) : new Pool(origin, opts);
+    return new Pool(origin, COPILOT_HOSTS.has(hostname) ? { ...opts, pipelining: 0 } : opts);
   },
 }));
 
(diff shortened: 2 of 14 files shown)


Menci:
## Conclusion on the `headers.delete('content-length')` change

This PR adds two `headers.delete('content-length')` calls in `copilotAuthedFetch` with the justification that *"interceptors may mutate the request body after content-length is set, causing `RequestContentLengthMismatchError`."* Reading undici v6's source shows this defense protects against a condition that cannot occur in this codebase.

### What actually throws `RequestContentLengthMismatchError`

In `undici/lib/dispatcher/client-h1.js` (v6.21.0) the error has exactly four throw sites — L908, L1162, L1265, L1340 — and every one of them is guarded by `request.contentLength !== null` (or the writer-local `contentLength` parameter being non-null, which is derived from the same value).

`request.contentLength` has exactly one writer, at `undici/lib/core/request.js:369-373`:

```js
} else if (request.contentLength === null && headerName === 'content-length') {
  request.contentLength = parseInt(val, 10)
  ...
}
```

It defaults to `null` (L150) and is only set when a `content-length` header is observed on the request. **No caller-supplied `content-length` header in `init.headers` ⇒ `request.contentLength` stays `null` ⇒ none of the four throw sites can fire on a header-math mismatch.**

### What undici does for a string body

For `body: JSON.stringify(...)`, the WHATWG `extractBody` algorithm at `undici/lib/web/fetch/body.js:85-91, 214-216` keeps `source = object` and computes `length = Buffer.byteLength(source)`. The fetch driver at `undici/lib/web/fetch/index.js:1410, 1426-1434` then reads that `length` and appends it as the `content-length` header on the outgoing request:

```js
const contentLength = httpRequest.body ? httpRequest.body.length : null   // L1410
...
httpRequest.headersList.append('content-length', contentLengthHeaderValue, true)   // L1434
```

The bytes that actually get written are `textEncoder.encode(source)` (body.js:56), whose length is the same `Buffer.byteLength(source)`. The header value and the byte stream are derived from one source — they cannot disagree.

One subtlety worth noting: `HeadersList.append` (`undici/lib/web/fetch/headers.js:166-183`) appends rather than overrides, joining with `, `. So if a caller *did* place `content-length` into `init.headers`, the result on the wire would be `caller-value, undici-computed-value`, and the L908 strict check would fire. This is the only realistic header-math path to the error — and it requires the caller to have supplied `content-length` in the first place.

### Why the Floway codepath cannot trigger this

For Copilot, the body passed to `copilotAuthedFetch` is always a freshly-`JSON.stringify`'d string built from the post-interceptor payload (`packages/provider-copilot/src/provider.ts:180, 202, 354`). Interceptors *do* mutate `ctx.payload` (force-store-false, vendor-{deepseek,qwen}-normalize, server-tool-shim, web-search-shim, strip-image-generation, etc.), but the stringify happens *after* the interceptor chain runs, so the byte length always matches the post-mutation payload.

The headers passed to runtime fetch are built in `packages/provider-copilot/src/auth.ts:243-274`. They start from `new Headers(init.headers)` where `init.headers` is `undefined` at every call site, then add a fixed pinned set plus a merge from `options.headers` (= `ctx.headers`). A workspace-wide grep finds **zero writers** of `content-length` into either `ctx.headers` or anywhere on the request-direction path. So no caller-supplied `content-length` ever reaches undici, `request.contentLength` stays `null`, and the strict check is structurally unreachable.

The userspace proxy path doesn't change this — `packages/http/src/fetch-on-stream.ts:91, 101` already strips any caller-supplied `Content-Length`/`Transfer-Encoding`/`Connection` and recomputes from actual body bytes, so even with a proxy in play there is no mismatch source.

### What the PR author actually observed

The four stack traces in undici issues [#2543](https://github.com/nodejs/undici/issues/2543), [#2898](https://github.com/nodejs/undici/issues/2898), [#3288](https://github.com/nodejs/undici/issues/3288), and [#2104](https://github.com/nodejs/undici/issues/2104) all surface from `AsyncWriter.end` / `writeIterable` — i.e. client-h1.js **L1340**, the `bytesWritten !== contentLength` end-of-write check. In every one of those scenarios, `contentLength` was computed by undici itself from the body, and the bytes that *would* have been written were that same body. The mismatch arises because **the socket is destroyed mid-write** — by a server-initiated keep-alive close, an aborted redirect, or a failed retry of a non-replayable body.

This is the same root cause as the `UND_ERR_SOCKET: other side closed` symptom this PR also fixes. \`api.individual.githubcopilot.com\` closes the keep-alive socket immediately after one exchange; depending on whether the close lands before or during the next request's body write, undici surfaces the race as either `other side closed` or `RequestContentLengthMismatchError`. The `pipelining: 0` change kills HTTP/1.1 connection reuse entirely and therefore resolves both symptoms. The header `delete` does nothing for either, because L1340 fires from a value undici computed itself — stripping inbound `content-length` from a header set that never contained it has no effect on the writer's `contentLength` parameter.

### Decision

Both `headers.delete('content-length')` calls in `copilotAuthedFetch` are dead defensive code:
- They cannot fire today (the header is never set there).
- They would not have helped against the actual error class this PR observed (which fires from socket-destruction during write, not from header math).
- Per the "no fake robustness" rule, remove them rather than keep guards against impossible states.

The dispatcher change (narrowed to the Copilot host) is the only piece of the connection-side fix that does real work.

PR #90, opened 2026-06-22
fix(responses): recognize compaction trigger items

M1k0t0:
## Summary

Recognize `compaction_trigger` as a first-class Responses item type when formatting stored response item ids.

Without this prefix mapping, responses that include compaction trigger items can hit the generic unknown-item path instead of receiving a stable item id prefix. This adds the `cmpt` prefix and covers it in the item formatting tests.

## Reproduction

Use the `/azure-api.codex` endpoint and continue a conversation until automatic compaction is triggered.

Observed behavior: once auto compaction is triggered, the request fails instead of continuing normally.

Screenshot: 

IMG_6491

### DIFF
--- a/packages/gateway/src/data-plane/llm/responses/items/format.ts
+++ b/packages/gateway/src/data-plane/llm/responses/items/format.ts
@@ -22,6 +22,7 @@ const itemTypePrefixes = {
   tool_search_call: 'ts',
   tool_search_output: 'tso',
   compaction: 'cmp',
+  compaction_trigger: 'cmpt',
   image_generation_call: 'ig',
   code_interpreter_call: 'ci',
   local_shell_call: 'lsh',
--- a/packages/gateway/src/data-plane/llm/responses/items/format_test.ts
+++ b/packages/gateway/src/data-plane/llm/responses/items/format_test.ts
@@ -22,6 +22,7 @@ const explicitPrefixes = [
   ['tool_search_call', 'ts'],
   ['tool_search_output', 'tso'],
   ['compaction', 'cmp'],
+  ['compaction_trigger', 'cmpt'],
   ['image_generation_call', 'ig'],
   ['code_interpreter_call', 'ci'],
   ['local_shell_call', 'lsh'],


Menci:
Closing in favor of #93 (merged). The throw you reported is real — quick summary of why #93 went a different direction.

## Test

Local wrangler dev against prod Copilot upstream (`gpt-5.5`). Codex-shaped client: turn 1 = plain user message → save `response.id`; turn 2 = `previous_response_id` + `input: [{ type: 'compaction_trigger' }]`. Read the persisted snapshot back from D1 after turn 2.

Three branches, same test:

| branch | turn 2 HTTP | snapshot `itemIds` for turn 2 |
|---|---|---|
| `main` | **502** | (request died before snapshot) |
| this PR | 200 | 8 items: 5 prior-history rows + `cmpt_*` (`compaction_trigger`, origin=input) + 2× `cmp_*` (`compaction`) |
| #93 | 200 | 2 items: 2× `cmp_*` (`compaction`) only |

`main` stack:
```
TypeError: Unknown Responses item type: compaction_trigger
  at prefixForItemType
  at createStoredResponsesItemId
  at LayeredStatefulResponsesStore.stageInputItem
  at stageUserInputItems
  at prepareResponsesServePlan
  at responsesServe.generate
```

## Why a prefix isn't enough

Adding `cmpt` does silence the throw, but the resulting `append`-mode snapshot keeps every prior message + reasoning row alive next to the compaction blob — so the next turn that resolves this `previous_response_id` re-streams the entire history we just paid the upstream to summarize away, exactly the thing compaction exists to avoid. The trigger itself also ends up persisted as an input row, which has no reader: per [codex's own `models.rs`](https://github.com/openai/codex/blob/9f06cf1a0907f650887ca26af5cee13b4f9a13ea/codex-rs/protocol/src/models.rs#L1122-L1128) the Responses API rejects an `id` on `compaction_trigger`, codex never re-sends it, and its rollout/history policy filters it out entirely.

#93 treats the trigger as a control signal at the storage seam: early-return in `stageInputItem` so no row is minted, and force `snapshotMode: 'replace'` (the same mode `/responses/compact` already uses) for any generate whose input carries a trigger. That fixes both the throw and the snapshot semantics in one shot.

## Side note worth flagging

One detail that surfaced during testing — this PR also leaves a permanent `cmpt_*` row in D1 (dedup by `content_hash` means just one ever exists, but it never expires). Once that row is there, `main` *also* stops throwing on the same workspace, because `stageInputItem`'s `contentHash` branch reuses the existing row before ever reaching `createStoredResponsesItemId`. So if any workspace ran this PR's code once, the `main` throw becomes invisible there without actually being fixed. I had to manually `DELETE` that row to reproduce 502 on `main` for the comparison above.

Thanks for the repro and the report — it was a real bug, just with a wider blast radius than the prefix alone covers.

PR #91, opened 2026-06-22
fix(responses): preserve Codex-compatible response chaining

M1k0t0:
## Summary

Preserve response chaining for Codex-compatible clients across Floway's Responses surfaces.

The README documents that `store:false` on a Responses WebSocket turn should disable durable persistence, but should still keep response state inside the current socket session so later messages on the same WebSocket can use `previous_response_id`.

Before this change, `store:false` could skip snapshot recording entirely, so a follow-up message in the same socket could fail to resolve a response id that Floway had just returned.

This change makes WebSocket `store:false` session-local instead of stateless: it does not write durable response state, but it keeps enough in-memory socket state for same-session chaining to work.

## Codex Compatibility

Codex under `/azure-api.codex` normally sends `store:true`, but users may also point Codex-compatible clients at `/v1/responses`. In that path, strict handling of `store:false` can make chaining brittle even though Floway has enough local context to preserve the expected client behavior.

Floway returns its own response ids, so this change preserves replay state for Codex-style requests where needed.
That keeps follow-up turns usable across `/azure-api.codex` and `/v1/responses` without changing the upstream wire body.

## What Changed

- Keep WebSocket `store:false` snapshots in the socket-local session store.
- Preserve durable persistence for `store:true` and omitted `store`.
- Add Codex client detection for `/v1/responses` chaining.

## Tests

- `corepack pnpm vitest run --project @floway-dev/gateway packages/gateway/src/data-plane/codex/routes_test.ts packages/gateway/src/data-plane/llm/responses/http_test.ts packages/gateway/src/data-plane/llm/responses/websocket_test.ts`
- `corepack pnpm --filter @floway-dev/gateway run typecheck`
- `corepack pnpm eslint packages/gateway/src/data-plane/codex/client-detection.ts packages/gateway/src/data-plane/codex/routes_test.ts packages/gateway/src/data-plane/llm/responses/http.ts packages/gateway/src/data-plane/llm/responses/http_test.ts packages/gateway/src/data-plane/llm/responses/items/store.ts packages/gateway/src/data-plane/llm/responses/websocket.ts packages/gateway/src/data-plane/llm/responses/websocket_test.ts`

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,15 +129,17 @@ the shim driving the internal multi-turn loop and replaying prior
 `/v1/responses` stores replayable Responses input and output items for API-key
 scoped HTTP requests. Clients can send `previous_response_id` to continue from
 a stored snapshot, or resend full input history; repeated full-history input is
-deduplicated by content hash instead of stored again. HTTP `store: false` does
-not create durable snapshots or input payload rows, but it keeps output item
-metadata for routing; if a later `store: true` request echoes that item with a
-full payload, the metadata row is filled in place.
+deduplicated by content hash instead of stored again. For ordinary clients, HTTP
+`store: false` does not create durable snapshots or input payload rows, but it
+keeps output item metadata for routing; if a later `store: true` request echoes
+that item with a full payload, the metadata row is filled in place.
 
 The same endpoint accepts `GET` WebSocket upgrades for streaming Responses
-events. WebSocket `store: false` keeps replay state only inside the open
-session, so same-socket `previous_response_id` works without writing those
-items or snapshots to durable storage.
+events. For ordinary clients, WebSocket `store: false` keeps replay state only
+inside the open session, so same-socket `previous_response_id` works without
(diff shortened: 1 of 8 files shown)


Menci:
Codex sends `store: false` over HTTPS Responses API and doesn't expect the server to *presist* the response and items. Codex only sends `previous_response_id` in a WebSocket Responses API session, utilizing session-local response and items storage.

M1k0t0:
Thanks, I verified this again and agree with the HTTP part: Codex over HTTPS Responses API sends `store:false` and does not rely on the server persisting responses/items there.

I’ll downscope the PR accordingly. The remaining issue is only the WebSocket behavior documented in the README: `store:false` should not durable-store the response, but response/items should still be available inside the same WebSocket session so `previous_response_id` works session-locally. 

So I’ll remove the HTTP Codex-compatible persistence part from this PR and keep it focused on WebSocket session-local storage only.

PR #97, opened 2026-06-23
fix(provider-codex): sanitize Responses message ids before dispatch

M1k0t0:
## Summary

- strip non-OpenAI message ids from Codex Responses input before upstream dispatch
- preserve valid `msg*` message ids and leave non-message item ids untouched
- handle both typed message items and typeless easy input messages when they reach the Codex provider boundary

## Why

Codex can echo local/review rollout message ids that the upstream Responses API rejects. Those ids are client-local, not upstream-owned, so Floway should remove them before sending the request to Codex upstream while preserving the message content.
This PR is scoped to the Codex provider wire boundary. A separate gateway-side normalization fix is still needed for stored native Responses requests where typeless `{ role, content }` messages are staged before provider interceptors run.

## Reproduction

1. Run `/review` and wait for the review to complete.
2. Send another message in the same Codex conversation.

### DIFF
--- a/packages/provider-codex/src/interceptors/responses/index.ts
+++ b/packages/provider-codex/src/interceptors/responses/index.ts
@@ -4,14 +4,15 @@
 
 import { injectDefaultInstructions } from './inject-default-instructions.ts';
 import { injectSessionId } from './inject-session-id.ts';
+import { sanitizeMessageIds } from './sanitize-message-ids.ts';
 import { stripUnsupportedFields } from './strip-unsupported-fields.ts';
 import type { ResponsesBoundaryCtx } from './types.ts';
 import type { Interceptor } from '@floway-dev/interceptor';
 
-// Order rationale: none of the three interceptors below read or write a field
-// the others touch, so order is positional only. inject-session-id last is
+// Order rationale: none of the interceptors below read or write a field the
+// others touch, so order is positional only. inject-session-id last is
 // conventional but not load-bearing — it hashes only `instructions + first
-// user-message text`, neither of which is mutated by the other two.
+// user-message text`, neither of which is mutated by the earlier mutators.
 //
 // Each interceptor is generic over the terminal result type: the streaming
 // `/responses` chain runs to ProviderStreamResult, the compaction chain runs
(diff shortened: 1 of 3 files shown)


Menci:
But, how does it work on Codex official client+server?

M1k0t0:
I dug into this more. The relevant official Codex behavior has two paths.

For normal Codex Responses requests, the client strips item ids when the `item_ids` feature is off, and that feature is disabled by default. Therefore the official Codex client does not normally send local rollout ids such as `review_rollout_user`.

However, when Codex thinks it is connected to an Azure Responses endpoint, for example because the base URL contains `azure-api.`, it sets `store: true`, and the id-stripping path is skipped when `store` is true. After that, the official Codex CLI really does send the local `/review` rollout ids to Floway even though `features.item_ids` is not enabled.

Floway then forwards to the Codex upstream, where those local ids are not valid Codex message ids. This patch handles that compatibility boundary: preserve upstream-shaped message ids (`msg...`) but drop client-local message ids like `review_rollout_user` before forwarding to Codex.

This PR is still draft for a separate reason: for stored native Responses requests, Floway’s gateway can currently fail while staging typeless `{ role, content }` message inputs before the request reaches the Codex provider boundary. I'd like to fix that upper gateway normalization path first, then come back and finalize this provider-level sanitizer change.

Refs:

- Guard that skips id stripping when `item_ids_enabled || store`:  
  [codex-rs/core/src/client.rs#L871-L878](https://github.com/openai/codex/blob/24423f57124ceba1e1ae50d808caaf991c99b479/codex-rs/core/src/client.rs#L871-L878)

- `store` is set from `provider.is_azure_responses_endpoint()`:  
  [codex-rs/core/src/client.rs#L852-L860](https://github.com/openai/codex/blob/24423f57124ceba1e1ae50d808caaf991c99b479/codex-rs/core/src/client.rs#L852-L860)

- `azure-api.` is one of the Azure URL markers:  
  [codex-rs/codex-api/src/provider.rs#L116-L126](https://github.com/openai/codex/blob/24423f57124ceba1e1ae50d808caaf991c99b479/codex-rs/codex-api/src/provider.rs#L116-L126)

PR #99, opened 2026-06-24
For local Node.js deployment, set environment variable RUNTIME_LOCATION=local

m13253:


cubercsl:
This regression may be introduced in 34c824c. 
I think a better solution is to replace previous `getEnv('RUNTIME_LOCATION')` with `getEnvOptional`.

m13253:
> I think a better solution is to replace previous `getEnv('RUNTIME_LOCATION')` with `getEnvOptional`.

Menci rejected #98.

PR #92, opened 2026-06-22
fix(provider-codex): match CLIProxyAPI response headers

M1k0t0:
Match the Codex Responses request headers used by CLIProxyAPI for `/backend-api/codex/responses`.

We hit a case where the same upstream account and same proxy egress IP behaved differently depending on the client shape:

- `/codex/models` could fetch and list models normally.
- Codex CLI worked normally from a machine using the same egress IP.
- Floway requests to `/backend-api/codex/responses` still returned upstream `403`.

That pointed away from account validity, token refresh, or proxy reachability, and toward request fingerprint differences on the Responses data-plane call.

The captured 403 body starts with a normal HTML document:
```html
<html>
  <head>
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <style global>body{font-family:Arial,Helvetica,sans-serif}.container{align-items:center;display:flex;flex-direction:column;height:100%;justify-content:center;width:100%}@keyframes enlarge-appear{0%{opacity:0;transform:scale(75%) rotate(-90deg)}to{opacity:1;transform:scale(100%) rotate(0deg)}}.logo{color:#8e8ea0;padding-bottom:1.5rem}.scale-appear{animation:enlarge-appear .4s ease-out}@media (min-width:768px){.scale-appear{height:48px;width:48px}}.data:empty{display:none}.data{border-radius:5px;color:#8e8ea0;max-width:420px;text-align:center;width:100%}.data ul{list-style:none;padding:0}.blocked-icon{color:#ef4444}.message{align-items:center;display:flex;font-size:1.5rem;gap:1rem;justify-content:center}.explanation{font-size:.9rem;line-height:1.5;max-width:420px;opacity:50%;text-align:center;width:100%}.explanation a{color:#000}@media (prefers-color-scheme:dark){body{background-color:#343541;color:#eee}.logo{color:#acacbe}.explanation a{color:#fff}}</style>
  </head>
  <body>
    <div class="container">
      <div class="logo">
        <svg
          width="41"
          height="41"
          viewBox="0 0 41 41"
          fill="none"
          xmlns="http://www.w3.org/2000/svg"
          strokeWidth="2"
          class="scale-appear"
        >
          <path
[...]

PR #107, opened 2026-06-24
fix(provider-codex): sync Codex Responses request fingerprint with the official CLI

M1k0t0:
Follow-up to #92.

# Context

#92 narrowed the Codex Responses 403 to request fingerprint differences: `/codex/models` worked, the official Codex CLI worked from the same egress path, but Floway `/backend-api/codex/responses` still received upstream HTML 403s.

This PR continues that work by aligning Floway’s Codex Responses request shape with the official `openai/codex` Rust client instead of only matching another relay implementation.

Thanks to @Menci for the detailed #92 review, the follow-up items called out there have been addressed in this PR.

# Changes

- Synthesize Codex-style session/thread/turn/window identity for `/codex/responses`.
- Send canonical `session-id`, `thread-id`, `x-client-request-id`, `x-codex-window-id`, `x-codex-turn-metadata`, `client_metadata`, and `prompt_cache_key` using the official Codex semantics.
- Persist Codex session/window metadata in Responses snapshots so replay, forks, and compaction keep the right upstream thread/window scope.
- Normalize `session_id` to canonical `session-id` at the provider boundary.
- Scrub downstream-supplied Codex metadata headers before provider dispatch and replace them with Floway-owned metadata.
- adds `responses_snapshots.metadata_json` via D1 migration
- Add snapshot metadata storage for Responses snapshots.
- Update the Codex data-plane identity to the current official CLI version/UA shape.
- Remove the uncited `~88% cache hit` comment while rewriting session-id injection, replacing it with qualitative Codex session/cache scope wording. Note: the `~88% cache hit` comment called out in #92 was already present on upstream/main before this branch.

# Verification

- `pnpm vitest run --project @floway-dev/provider-codex packages/provider-codex/src/fetch_test.ts packages/provider-codex/src/compaction_test.ts packages/provider-codex/src/interceptors/responses/inject-session-id_test.ts`
- `pnpm vitest run --project @floway-dev/gateway packages/gateway/src/data-plane/llm/responses/serve_test.ts packages/gateway/src/data-plane/llm/responses/items/store_test.ts packages/gateway/src/repo/responses-items_test.ts packages/gateway/src/data-plane/shared/inbound-headers_test.ts`
- `pnpm --filter @floway-dev/provider-codex run typecheck`


packages/provider-codex/src/responses-state.ts
Menci, line 1: I don't want to allow some concepts of ONE provider to the whole gateway.

M1k0t0:
I agree with the concern, and I changed the implementation to keep Codex-specific concepts out of the gateway.

In the latest version, the gateway no longer knows about Codex session ids, turn ids, account ids, or Codex-specific metadata keys. Those are all owned by `provider-codex`.

The gateway-side changes are now limited to provider-neutral lifecycle hooks:

- `prepareResponsesRequest`: lets a provider adjust the upstream Responses request before dispatch.
- `beforeResponsesSnapshotCommit`: lets a provider attach continuation metadata before a terminal snapshot is committed.

PR #108, opened 2026-06-25
feat(copilot): expose Anthropic Fast Mode through Copilot upstream

Menci:
## Summary

Bridge Anthropic's [Fast Mode](https://docs.claude.com/en/build-with-claude/fast-mode) (request-side `speed: "fast"`, response-side `usage.speed: "fast"`) over GitHub Copilot's upstream, entirely inside the Copilot provider. Copilot exposes Fast Mode as a separate raw model id (`claude-opus-4.6-fast` today; `claude-opus-4.8-fast` succeeds it 2026-06-29) but does not speak the `speed` field on either side of the wire — the gateway translates both directions so Anthropic-compatible clients see one canonical surface.

### Behavior

The `claude-opus-4.6-fast` raw variant collapses into the public id `claude-opus-4-6` in `/v1/models`, the same way `-1m` / `-xhigh` already do. A Messages request with `speed: "fast"` routes to the `-fast` raw variant, the field is stripped from the upstream wire body, and `usage.speed = "fast"` is stamped onto every outbound `message_start` / `message_delta` frame so downstream sees the marker the billing path and Anthropic-compatible clients expect.

Per Anthropic's no-silent-fallback contract, `speed: "fast"` on a model with no `-fast` raw variant short-circuits with HTTP 400 `invalid_request_error` before any boundary interceptor or upstream call runs. The error string (`'<model-id>' does not support the \`speed\` parameter.`) is byte-identical to the wire string Anthropic emits on the real API — recovered from a [regression test in Yeachan-Heo/gajae-code](https://github.com/Yeachan-Heo/gajae-code/blob/main/packages/ai/test/anthropic-fast-mode.test.ts) that records live Anthropic 400 bodies verbatim, with independent corroboration in three other open-source gateways.

Unknown `speed` values (`"priority"`, anything else) are forwarded to Copilot verbatim so the upstream owns the rejection, mirroring how Anthropic itself behaves with stray fields.

### Selection

A new `fast` hint dimension joins `context1m` and `reasoningEffort` in `model-selection.ts`. Selection itself stays best-effort (cascading narrows that roll back on miss); the hard contract is enforced at the `callMessages` entry point. This split means a unit caller exercising selection in isolation still gets a reasonable answer even when no `-fast` variant exists, while the production path 400s before reaching here.

### Pricing

Opus 4.6 and 4.7 carry the public Fast Mode rates (6× base = $30 in / $150 out per MTok); Opus 4.8 carries the GA rates (2× base = $10 / $50); Opus 4.5 keeps base-only (no `-fast` variant in Copilot's catalog). All under `tiers.fast` so `usage.speed = "fast"` → `tier = "fast"` resolves the override at billing time.

### Gateway latency-recorder contract

The `recordUpstreamLatency` contract that wraps each provider call was tightened along the way. `recorder.durationMs()` now returns `number | null` — `null` honestly means "the provider never wrapped a fetch" (a legitimate state on the request-side-400 short-circuit). The `requireRecordedDurationMs(recorder, label)` helper is the explicit assert for call sites that need a real number (every success-side record, plus the count_tokens contract guard), so the "you forgot to wrap" error stays attached to the line that depends on the value rather than firing from inside the recorder.

## Test plan

- [x] `pnpm run typecheck` — clean across the workspace
- [x] `pnpm run lint` — clean
- [x] `pnpm run test` — 3350 / 3350 passing
- [x] Manual end-to-end smoke against a Node mock-Copilot upstream (dev-only artifact, not committed): `speed: "fast"` routes to `claude-opus-4.6-fast`, response stamps `usage.speed: "fast"` on `message_start` and `message_delta`, request wire body has no `speed` field; `speed: "fast"` on `claude-haiku-4-5` short-circuits with the byte-identical 400; `speed: "priority"` (unknown) forwards verbatim
- [x] Byte-identical Anthropic Fast-Mode unsupported-model `error.message` confirmed against recorded live wire responses in third-party regression fixtures


Menci:
Updated based on review feedback:

1. **Verbatim Anthropic error string.** Found the actual wire string from a regression test in [Yeachan-Heo/gajae-code](https://github.com/Yeachan-Heo/gajae-code/blob/main/packages/ai/test/anthropic-fast-mode.test.ts) that records live api.anthropic.com 400 bodies verbatim, with independent corroboration in three other open-source gateway codebases. The 400 `error.message` now reads `'<model-id>' does not support the \`speed\` parameter.` (single-quoted model id, backticked parameter name, terminating period) — byte-identical to upstream.

2. **Relaxed the recordUpstreamLatency contract.** The previous `Promise.resolve()` workaround in the 400 short-circuit satisfied the strict "you must wrap a fetch" guard but lied to telemetry (claiming a 0 ms upstream call when none happened). Replaced it with a contract relaxation in the gateway-side helper: `providerStreamResultToExecuteResult` now takes the recorder and consumes `durationMs()` lazily — only the `ok=true` branch needs it. `ok=false` short-circuits flow through without ever triggering the guard. The other four direct call sites (count_tokens guard, responses-compact, image-gen, passthrough) keep the strict behavior because they genuinely consume the value.

Verification: `pnpm run typecheck`, `pnpm run lint`, and `pnpm run test` (3350 / 3350) are clean on `ce50a433`.

Menci:
Round-tripped on the contract again: `recorder.durationMs()` now returns `number | null` (no throw), and `requireRecordedDurationMs(recorder, label)` is the explicit assert helper for call sites that need the value. Semantics now match your framing:

- **ok=true**: success path always needs a real number → `requireRecordedDurationMs` at the use site.
- **ok=false**: latency is whatever the recorder happens to know, threaded into `recordUpstreamHttpFailure(ctx, ctx, durationMs)`. Today the failure branch still drops it (the `upstream_success` scope only carries the error counter), but the API is now honest so a future schema can record failure latency without touching call sites.

Direct consumers (count_tokens guard, responses-compact, image-generation, passthrough-serve) use `requireRecordedDurationMs` so the throw stays attached to the exact line that depends on the value. Performance unit tests updated to assert the new contract. typecheck / lint / test (3350/3350) clean on `ac0d1eb7`.

PR #121, opened 2026-06-26
fix(responses): cross-protocol callers use a no-op inner store

Menci:
## Summary

Cross-protocol Messages / Gemini / Chat-Completions callers route through `responsesAttempt.generate` for the inner Responses dispatch after translating the request shape. They previously forwarded their outer protocol's `StatefulResponsesStore` into that inner call, so the inner Responses path's item writes and snapshot commits landed under rows the outer protocol never reads back. The result was orphan rows accumulating per cross-protocol turn — invisible at runtime but a long-term storage cost the outer protocol doesn't own.

Add a `createNoOpResponsesStore(apiKeyId)` factory whose `reads` / `itemWrites` / `snapshotWrites` are empty so every backing-write iteration is a no-op. Pass it to the inner Responses call in place of the forwarded outer store. The three cross-protocol attempt files (`messages/attempt.ts`, `gemini/attempt.ts`, `chat-completions/attempt.ts`) each pick it up at exactly one site.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm run test`
- [x] Tests assert each cross-protocol attempt path's inner Responses call receives a no-op store (no items / snapshots committed under the outer protocol's apiKeyId)

### DIFF
--- a/packages/gateway/src/data-plane/chat/chat-completions/attempt.ts
+++ b/packages/gateway/src/data-plane/chat/chat-completions/attempt.ts
@@ -3,7 +3,7 @@ import type { ChatCompletionsInvocation } from './interceptors/types.ts';
 import { messagesAttempt } from '../messages/attempt.ts';
 import { responsesAttempt } from '../responses/attempt.ts';
 import { rewriteStoredResponsesItemsForCandidate } from '../responses/items/rewrite.ts';
-import type { StatefulResponsesStore } from '../responses/items/store.ts';
+import { createNoOpResponsesStore, type StatefulResponsesStore } from '../responses/items/store.ts';
 import { providerStreamResultToExecuteResult, buildUpstreamCallOptions } from '../shared/attempt-helpers.ts';
 import type { ProviderCandidate } from '../shared/candidates.ts';
 import { tryCatchChatServeFailure } from '../shared/errors.ts';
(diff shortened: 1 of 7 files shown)


Menci:
Closing this — the premise is wrong.

I framed this as fixing "orphan rows": cross-protocol Messages / Gemini / Chat-Completions callers route through `responsesAttempt.generate` for the inner translated dispatch and (I claimed) write items + snapshots into the OUTER protocol's store that nobody reads back. Hence: pass a no-op store to the inner call.

Two problems with that:

1. **The outer store's `snapshotWrites` is already empty.** `createNonResponsesSourceStore` (`packages/gateway/src/data-plane/chat/responses/items/store.ts:575-582`) configures `snapshotWrites: []`. Cross-protocol's `commitSnapshot` call iterates an empty list — naturally a no-op. The `snapshotMode: 'none'` arg on the cross-protocol call site is belt-and-suspenders, not load-bearing. No snapshot writes were ever happening.

2. **Item writes are not orphan.** The outer store's `itemWrites` IS populated and `reads` IS populated, both pointing at the repo backing. The inner Responses dispatch writes items by content hash; the SAME store (same apiKeyId, same backing) is opened on a subsequent cross-protocol turn from the same client; the gateway looks those items up by content hash for cross-turn affinity routing. The writes are how cross-turn affinity stays bound to the same upstream (and any upstream-bound encrypted reasoning blobs keep working).

Swapping in a no-op `reads: [], itemWrites: [], snapshotWrites: []` store kills that affinity. Turn 2 of a Messages-via-Responses conversation would find nothing for its content-hash query, dispatch routing falls back to general selection, and a previously upstream-pinned context loses its binding.

Closing without merging. The snapshotMode override field cleanup (the work that originally motivated thinking about this) can land standalone — cross-protocol callers can drop their now-redundant `snapshotMode: 'none'` arg because the store's empty `snapshotWrites` does the equivalent work, and no orphan-rows / no-op-store rearrangement is needed.

PR #123, opened 2026-06-28
refactor(resolve): rework the model resolution pipeline

Menci:
## Summary

Reworks how the gateway turns an inbound `model` string into a candidate
the dispatch layer can call. Reference doc:
[`RESOLUTION.md`](./RESOLUTION.md) (added in this PR).

The pipeline reads as three orthogonal concerns:

- **Catalog assembly** — every enabled upstream's SWR-cached catalog is
  merged by public id into `CatalogModel[]` (public metadata) +
  `Map<publicId, Provider[]>` (reverse index). The metadata view and the
  dispatch view are two artefacts, not one struct with five copy fields.
- **Resolution** — one shared function,
  `enumerateModelCandidates({upstreamIds, model, kind, scheduler, currentColo})`,
  fans the inbound id across the visible upstreams' addressable surfaces,
  retries once with an `-YYYYMMDD` suffix stripped when nothing matched,
  and keeps only resolutions whose `model.kind === kind`. Chat,
  embeddings, images, and `/v1/completions` all go through it.
- **Endpoint selection** — pushed out of resolution into the attempt layer.
  Each chat protocol owns a `chatTargetPicker(preference)` factory whose
  `canServe` filters the candidate list and whose `pick` decides which
  upstream wire the request goes out on. Passthrough endpoints declare a
  `modelServesEndpoint` callback and iterate matches the same way they
  did before.

The provider-side `resolveRequestedModelId` hook (claude-* dated-id remap
on Copilot, dated-id remap on Claude Code) was a per-provider compatibility
shim. It collapses into a single gateway-level `-\d{8}$` retry that runs
through the same full resolve pipeline, so dated ids now compose cleanly
with the rest of the resolver. Variant suffixes the Copilot hook also
stripped (`-high`/`-xhigh`/`-1m`/`-fast`) are no longer accepted at the
gateway boundary — clients address those by the public id plus the
matching request-body field, which is the documented contract.

## Naming and layering pass

Builds on top of the resolution rework with a series of targeted renames
so the names match what each piece actually carries. Reads better, gives
the upcoming alias feature a clean surface to plug into:

- `chat/responses/items/` -> `chat/items/`. The stored-items subsystem
  (store, affinity walk, output wrap, item rewrite, normalize-assistant-content)
  is shared across messages / gemini / chat-completions / responses, not
  responses-only.
- Per-request stored-items `store` moves into a new `ChatGatewayCtx`
  (extends `GatewayCtx` with `store: StatefulResponsesStore`), constructed
  by each chat HTTP/WS entry via `createChatGatewayCtxFromHono(c, opts,
  storeFactory)`. Drops `store` from every chat serve/attempt/narrow args
  object and from `ResponsesInvocation`. Passthrough endpoints stay on
  plain `GatewayCtx`.
- `routing` -> `narrow-by-item-affinity`. The function does not pick a
  route; it narrows the candidate list and reorders by stored-item
  affinity. `narrowMessagesByItemAffinity` / `narrowResponsesByItemAffinity`
  / `narrowGeminiByItemAffinity` / `narrowChatCompletionsByItemAffinity`,
  returning `readonly ModelCandidate[] | ChatServeFailure` directly (no
  wrapper).
- Chat-target picker collapses into a single `chatTargetPicker(preference)`
  factory returning `{ canServe, pick }`. `canServe` is used by serve to
  filter viable candidates; `pick` is used by attempt and is non-null —
  serve has already filtered, so picking a viable candidate is an
  invariant, not a fallible step.
- `UpstreamRecord.provider` -> `.kind`. The field was named `provider` but
  typed `UpstreamProviderKind`; align the field name with what it
  actually carries. Renames the wire DTO discriminator too. SQL column
  stays `provider`; sql.ts adapts between the two.
- `ModelProviderInstance` (outer) -> `Provider`; `ModelProvider` (inner)
  -> `ProviderInstance`; the inner field `provider:` -> `instance:`. The
  outer was a longer name than what it carries, and the inner field name
  collided with the outer (`candidate.provider.provider.callX`).
  Dispatch now reads as `candidate.provider.instance.callMessages(...)`.
- `ProviderCandidate` -> `ModelCandidate`. Originally a provider had a
  single model so "provider candidate" read as "a candidate provider for
  this request." Now a provider's catalog can emit multiple models, and
  the candidate carries `(provider, model, fetcher)` — it's a candidate
  model from a provider, not a candidate provider. Rename siblings to
  match: `ProviderModelResolution` -> `ModelResolution`,
  `enumerateProviderCandidates` -> `enumerateModelCandidates`,
  `resolveModelForProvider` -> `findModelInProvider`,
  test stub `stubProviderCandidate` -> `stubModelCandidate`.

## Behavioral changes

- Inbound id with no dated suffix that matches nothing -> 404 (unchanged).
- Inbound id with `-YYYYMMDD` that matches nothing -> 404 only after the
  stripped id also matches nothing (new).
- Inbound id that resolves but `model.kind` is wrong for the endpoint ->
  now 400 "does not support the X endpoint" (previously 404
  "is not available"). The id is known, just the wrong family.
- Inbound `claude-*-high` / `-xhigh` / `-1m` / `-fast` raw variants ->
  404 unless the upstream lists them verbatim (previously normalized by
  the Copilot hook).
- API wire shape: `UpstreamRecord` (POST/PATCH bodies, list response,
  data-transfer import/export, `/api/upstream-options`) now uses `kind`
  for the provider-kind discriminator instead of `provider`. Dashboard
  consumers updated. SQL column unchanged.

## Test plan

- [x] `pnpm run typecheck` — all packages
- [x] `pnpm run lint`
- [x] `pnpm run test` — 3518/3518 across 309 test files
- [ ] Manual smoke: chat / embeddings / images / completions / Gemini through a local upstream
- [ ] Manual: dated id (`claude-sonnet-4-5-20251001`) resolves to base on Claude Code upstream
- [ ] Manual: variant id (`claude-opus-4-7-xhigh`) now 404s on Copilot (intended; covered above)


Menci:
Superseded by the cascading split #129 → #130 → #131 → #132 → #133 → #134 (and follow-ups: inline pick.ts, canonicalize Responses input, iterate candidates — coming next).

PR #130, opened 2026-06-28
cleanup(resolve): retire stale 'binding' narration and other resolver-layer docs polish

Menci:
Follow-up cleanups against the resolver layer's prose and tests.

- Drop migration-only prose ('used to be X, now Y') from `RESOLUTION.md` Candidate Shape and inline comments.
- Retire the lingering 'binding' vocabulary in narration and one candidate-test name — the shape is `ModelCandidate { provider, model, fetcher }`, not a binding.
- Clarify retry-vs-kind-filter ordering in two comments.
- Trim restating prose from four resolver-layer files.
- Fix trailing-whitespace patching traces left in six test mocks.
- Dedupe a duplicate `stubUpstreamModel` and finish a half-rewritten comment.
- Refresh one comment that still narrated the pre-kind-filter flow.

No code behavior changes — only comments, names, and test scaffolding.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm run test`

### DIFF
--- a/RESOLUTION.md
+++ b/RESOLUTION.md
@@ -120,19 +120,24 @@ Steps:
    normalization hook above the catalog; an upstream that wants the
    gateway to accept a non-listed alternate id must list it in its
    catalog.
-4. **Kind-filter** the resolved set: keep only entries whose
+4. If the resolved set is empty **and** the inbound id matches
+   `-\d{8}$`, strip the eight-digit dated suffix and rerun steps 2–3
+   once against the stripped id. The retry is bounded to one iteration;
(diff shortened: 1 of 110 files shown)


Menci:
Folded into #129 — the cleanup commits were polish on top of #129's old patching-trace history. After #129 was rewritten as one cohesive commit, every relevant cleanup (binding-narration retirement, RESOLUTION.md migration prose, trailing-whitespace patching trace) was applied directly to #129's final state. Closing to keep the stack lean.

PR #131, opened 2026-06-28
refactor(chat): require failedUpstreams on ChatServeFailure model-* variants

Menci:
\`ChatServeFailure\` carries a \`failedUpstreams\` field on the \`model-missing\` / \`model-unsupported\` variants — the upstream names whose catalog fetch threw during this resolution. Promote it from optional to required so the renderer always has the list and the per-protocol failure copy can surface it consistently. Callers that have no failed upstreams pass an empty array; the construction sites that used to omit the field now construct it with \`failedUpstreams: []\` explicitly.

## Test plan

- [x] \`pnpm typecheck\`
- [x] \`pnpm lint\`
- [x] \`pnpm test\`

### DIFF
--- a/packages/gateway/src/data-plane/chat/shared/errors.ts
+++ b/packages/gateway/src/data-plane/chat/shared/errors.ts
@@ -3,8 +3,8 @@
 // carries the upstream names whose catalog fetch threw during this
 // resolution — surfaced parenthetically so the caller can tell a genuine
 // "no upstream has this model" miss from a transient outage where the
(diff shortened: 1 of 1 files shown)


Menci:
Closing — #129 absorbed all substantive changes (the required-field tightening on ChatServeFailure model-* variants and the appendFailedUpstreams type narrowing landed as part of the resolution rework). After rebasing onto the new main, the only surviving diff was a 2-word comment polish that doesn't justify a standalone PR.

PR #132, opened 2026-06-28
refactor(chat): collapse routing into a bare candidate list and own pickers in each attempt

Menci:
Three commits that reshape how chat protocols pick a candidate from the resolved set:

1. Collapse the routing decision into a bare candidate list and own pickers in each attempt — drop the wrapper struct so the value flowing through is `readonly ModelCandidate[]`.
2. Split the chat-target picker into `canServe` (used by serve to filter viable candidates) + `pick` (non-null; used by attempt — serve has already filtered, so picking a viable candidate is an invariant).
3. Pick the candidate list around stored-item ownership — let the affinity walk reorder the candidate list directly instead of going through an extra routing-decision struct.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm run test`


Menci:
Superseded — all three commits (canServe+pick split, bare candidate list, attempt owns the picker) are now in #129 directly. Closing to keep the stack lean.

PR #143, opened 2026-07-01
fix(web): show zero-token models in token/cost usage charts

yyyr-p:
## Summary

An upstream provider can produce requests that record no usage — every token dimension and cost stored as `0` — while still incrementing the request count. Such a model appeared in the **Requests** view of the usage charts but vanished from the token/cost views (e.g. the default **Total Tokens** view), making it look as if the model had no activity at all.

The data does distinguish the two cases: a model with real-but-zero-token traffic keeps `requests > 0`, whereas a model with genuinely no usage never appears in the records at all. A group emptied purely by cross-filtering also drops to zero requests, since requests aggregate the same cross-filtered records.

## Changes

- Keep a model in the token/cost (non-percent) charts when its selected-metric values are all zero but the group served requests in the range. Cross-filter-emptied groups (requests also zero) are still dropped, so no inert flat lines appear for genuinely absent activity.
- Render such a zero-token series as a flat, hoverable line pinned at zero: give it its own stack group with no fill so it no longer rides invisibly along the top edge of the series stacked below it. It contributes nothing to
  the main stack, so real totals are unchanged.
- Relax the tooltip filter to keep a zero-valued row whenever the group actually served requests in that bucket, so hovering a zero-token model still reports its requests/cost.

Percent metrics (cache hit / cached rate) are unchanged: a ratio over zero tokens is undefined, so those models stay excluded there.

## Renderer constraint

The initial fix substituted `?? 0` at the aggregate map. A subsequent refactor (e0f8b83e, ef9f82e4) swapped that for `?? null` to preserve a data-layer distinction between "no record in this bucket" and "real zero-token record"; a follow-up (1ae1dfc1) folds `?? 0` back for non-percent metrics because chart.js's stacked line renderer needs a numeric value at every index — under `y.stacked: true` a null pulls neighbors down to the axis instead of stacking on top of the running total, `spanGaps: false` gaps the line and its area fill collapses to the axis, and `spanGaps: true` stitches internal nulls but does not extend the line past the first or last non-null point, leaving leading and trailing "no record" buckets unlit. The `no-record vs real-zero` signal survives via the tooltip filter and dataset filter reading `details.requests` directly.

Two small pieces of the refactor stay on merit — a `!` type assertion (the aggregate map pre-populates the outer entry per bucketKey) and a shared `stack: 'axis'` group for every zero-line dataset (each contributes 0 to its stack, so a single shared stack renders identically). The shared stack needs one guardrail: `order: -1` on zero-line datasets, so the axis-level flat line paints on top of any main-stack area whose bottom sits at `y=0` instead of relying on per-stack draw ordering that the shared group loses.

## Verification

- `pnpm --filter @floway-dev/web run typecheck` and `eslint` pass.
- Manually verified in the dashboard against a mocked zero-token model: it now appears in the Total Tokens and Input Tokens views as a flat line at zero spanning the full x-axis (including leading/trailing "no record" buckets), and hovering its bucket shows `Req N, Cost $0, Total 0`. A sparsely-sampled non-zero model renders as one continuous stacked series across the whole axis, not disjoint fill lobes. Tooltip hovering a no-record bucket shows no phantom row for either kind of series.

### DIFF
--- a/apps/web/src/pages/dashboard/usage.vue
+++ b/apps/web/src/pages/dashboard/usage.vue
@@ -406,13 +406,20 @@ const buildStackedConfig = (groupKey: 'keyId' | 'model'): ChartConfiguration<'li
   const entries = groupKey === 'keyId'
     ? keyChartEntries([...presentGroups], keyMetadataForTokenRecords(allRecords, data.value?.keys ?? []), data.value?.keys.map(k => k.id) ?? [...presentGroups])
     : modelChartEntries([...presentGroups]);
-  // Cross-filtering can leave a group with all-zero (or, for percent metrics,
-  // all-null) values. Drop those datasets outright — legend entry and line both
-  // gone — instead of rendering an inert flat line. The own-dimension hidden set
-  // is a separate, restorable user toggle and keeps its struck-through entry.
+  // A group can hold all-zero (or, for percent metrics, all-null) values under
+  // the current metric for two distinct reasons, and requests tells them apart:
+  // cross-filtering that emptied the group leaves requests at zero too (details
+  // aggregate the same cross-filtered records), whereas a group with real but
+  // zero-token traffic — e.g. an upstream that reports no usage — still has
+  // requests > 0. Keep the latter as a flat line at zero so it stays visible in
+  // the token/cost views; drop the former outright, legend entry and line both
+  // gone, instead of rendering an inert line for a group with no activity. Own-
+  // dimension hidden groups are a separate, restorable toggle kept struck-through.
+  // Percent metrics stay null-only: a ratio over zero tokens is undefined.
+  const hasRequests = (id: string) => bucketKeys.some(k => (details.get(k)?.get(id)?.requests ?? 0) > 0);
   const datasetEntries = entries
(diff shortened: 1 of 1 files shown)


yyyr-p:
@Menci Testing this branch against a mocked zero-token model surfaced a rendering issue with the `?? null` refactor: chart.js's stacked line renderer needs a numeric value at every index, and null buckets produced three visible symptoms — stack misalignment (`y.stacked: true` pulls neighbors to the axis on null), gapped lines and area fills (`spanGaps: false`), and unlit leading/trailing "no record" buckets (`spanGaps: true` doesn't extend past the first/last non-null point, leaving zero-token flat lines shorter than the axis).

Pushed a follow-up (1ae1dfc1) that folds `?? 0` back at the aggregate map for non-percent metrics, and recovers the `no-record vs real-zero` distinction via `details.requests` in the tooltip filter and dataset filter. Two pieces of the refactor stay on merit: the `!` assertion, and the shared `stack: 'axis'` group (each zero-line contributes 0, so a single shared stack renders identically). The shared stack loses the per-stack draw ordering the per-id version had, so zero-line datasets now carry `order: -1` to paint on top of any main-stack area whose bottom sits at `y=0`.

PR body updated with a "Renderer constraint" section covering the same reasoning.

PR #154, opened 2026-07-04
feat(cursor): Cursor provider (chat + Tab) with cross-request session reuse

yyyr-p:
## Summary

Add Cursor as a 7th provider kind (alongside `copilot` / `custom` / `azure` / `codex` / `claude-code` / `ollama`) using Cursor's own CLI OAuth flow, and expose two data-plane surfaces:

- **Chat Completions** over Cursor's `RunSSE` + `BidiAppend` dual channel — multi-turn native MCP tool calling, image input, reasoning-effort / Max-Mode routing, and privacy (ghost) mode.
- **`/v1/completions` bridge** exposing Cursor Tab (`StreamCpp` on the geo-routed `us-only.gcpp.cursor.sh` backend) as an OpenAI-shaped edit-prediction endpoint. Zed's `open_ai_compatible_api` prompt formats — Zeta 2.1 (V0318) markers, V0615 hashed-regions (cross-file jumps), plus FIM and plain-prefix fallback — are all detected from the prompt and rendered back.

The interesting review surface is not the provider code — that is a mechanical implementation of Cursor's protobuf protocol whose shape is dictated by Cursor's own CLI. The interesting review surface is the **new runtime stack** that had to land underneath the provider to make it work on Workers at all.

## Infrastructure: `DurableHttpSession` — a new platform primitive

Cursor's agent stream is not stateless. A turn opens a `RunSSE` server-stream on which the model produces `exec_request` frames whose results the client is expected to write back on a companion `BidiAppend` channel — **on the same conversation**. A gateway that handles the tool-result follow-up on a different HTTP request must keep the outbound `RunSSE` response socket alive across the two request lifecycles and route the follow-up to whichever isolate holds that socket. Every existing Floway provider is one-request-in / one-response-out and does not have this problem.

Solution: **`packages/platform/src/durable-http-session.ts`** — a new abstract runtime contract with two symmetric operations, `acquire(key, init?)` and `release()` / `discard()`. `init` present = open (dial the upstream, hold the response stream). `init === null` = resume attempt (return the held stream to whoever wins the single-flight claim). The gateway core reaches this contract via `getDurableHttpSession()` and stays runtime-neutral.

Two concrete impls ship in this PR:

- **`apps/platform-cloudflare/src/durable-http-session-do.ts`** — a Cloudflare **Durable Object**, `DurableHttpSessionDO`. This is Floway's first DO that owns a live outbound socket. The DO stays alive because the outbound connection keeps it alive (Cloudflare's 2026-06-19 outbound-connections-keep-DOs-alive runtime behavior — bounded by CF's 15-minute cap). An idle alarm via `ctx.storage.setAlarm` (default 5 min) evicts abandoned sessions before then; the alarm evicts regardless of consumer attachment (matches the node impl's behavior). Chunk queues use a head-index queue with a fused concat/tail-slice so the DO does no per-chunk allocation on hot turns.
- **`apps/platform-node/src/in-process-durable-http-session.ts`** — a single-process in-memory impl for `pnpm run dev:node`. Same contract; no cross-instance semantics.

The single-flight claim (so a concurrent follow-up cannot race the resume) is a D1-backed repo, `cursorSessions` in the gateway repo composition, keyed by a session id derived from `(upstreamId, apiKeyId, headers, messages)` (see `session-id.ts`). On resume, `claim()` atomically locks the row for `CLAIM_TTL_MS = 60_000`; a miss / busy / lost-socket returns null, and the caller cold-resumes with the full transcript flattened into a fresh open (Cursor has no server-side memory of the prior turn after socket loss).

**New `wrangler.example.jsonc` surface driven by this stack:**

- New DO binding `DURABLE_HTTP_SESSION` → class `DurableHttpSessionDO`.
- New DO migration `v2` (`new_sqlite_classes: ["DurableHttpSessionDO"]`). Append-only.
- `compatibility_flags: ["nodejs_compat"]` — required because `@reclaimprotocol/tls/webcrypto` (used by `packages/http`'s userspace TLS dialer) still `import { webcrypto } from 'crypto'`.
- `scripts/check-wrangler.ts` now enforces `wrangler.jsonc` mirroring the example **bidirectionally**, with `account_id` as the one allowlisted per-contributor exception. This is a real behavior change to the deploy gate — every value the example pins must appear in the real config, and vice versa.

**New D1 migrations:**

- `0048_cursor_provider.sql` — adds `'cursor'` to the `upstreams.provider` CHECK (rebuild + copy per repo pattern).
- `0049_cursor_sessions.sql` — new `cursor_sessions` table for the resume single-flight.
- `0050_cursor_drop_quotasnapshot.sql` — data migration. The initial import path wrote `quotaSnapshot: null` on every cursor account row; a later behavior-preserving cleanup dropped the field from the state schema **and** from the strict key allowlist. Without this migration, every legacy cursor row would fail `assertCursorUpstreamState` with `unexpected key 'quotaSnapshot'` and return 500. Idempotent.

## Transport choice: HTTP/1.1 on `api2.cursor.sh` (not HTTP/2 on `api5.*`)

Cursor's control-plane exposes two backend families with different HTTP versions on the wire. `api2.cursor.sh` serves `RunSSE` + `BidiAppend` over HTTP/1 — both **HTTP/1.1** and HTTP/1.0 are accepted; Floway speaks 1.1. The `api5` family (`agent.api5.cursor.sh` privacy / `agentn.api5.cursor.sh` non-privacy) is **HTTP/2-only** — a forced HTTP/1.1 handshake returns `Received HTTP/0.9`; the default ALPN negotiates h2 and returns 200. Floway targets **`api2.cursor.sh` over HTTP/1.1** on purpose:

- Workers `fetch` is HTTP/1.1-native and interops with Cloudflare's outbound-connections-keep-DOs-alive lifetime story. Switching to HTTP/2 would mean introducing an h2 library into the transport and re-proving DO socket lifetime under it.
- Floway's `@floway-dev/proxy` runs upstreams over operator-configured proxy chains (userspace TLS + WebSocket-tunnel over TCP, VLESS, `runProxiedRequest`). Every dialer speaks a raw duplex byte stream and layers HTTP/1.1 on top. HTTP/2 would require multiplexing support inside every dialer, or bypassing the proxy layer entirely (which we won't do — proxy-through-Workers is a first-class feature).
- The privacy (ghost) mode toggle on `api2` is a **header** (`x-ghost-mode: true|false`), so we can honor it without leaving the HTTP/1.1 endpoint. Domain-based privacy on the `api5` family is the only reason to consider h2 for now, and it doesn't buy anything the header path doesn't already give us.

The Cursor Tab bridge dials **`us-only.gcpp.cursor.sh`** — a separate geo-routed backend, HTTP/1.1-compatible, Connect-proto server-stream (not the RunSSE dual channel). Same transport story.

## How Floway's cursor provider differs from other cursor gateways

Several open-source projects have already reverse-engineered pieces of Cursor's protobuf protocol; we cross-checked wire shapes and behavior traps against them while building this provider. Floway-specific choices worth calling out:

- **Cross-request session survival on serverless Workers.** No other cursor gateway we surveyed targets Workers, and none has a cross-request tool-result continuation story. This is the entire reason `DurableHttpSession` had to be invented.
- **HTTP/1.1 + operator-proxy compatibility over vendor-optimal transport.** Every other gateway we looked at is bare `fetch` with no proxy layer; the h2/h1 choice is invisible to them. For us it is load-bearing.
- **Inline history flatten into `UserMessage.text`, empty `turns[]`.** Native `turns` reference blobs the server can't fetch on a cold conversation ("Blob not found"). Prior tool rounds are framed explicitly (`[Assistant] → called foo(...)`, `[Tool result: foo] …`) so the model reads them as history without needing a blob dereference.
- **System prompt inlined into the flattened user text, tagged `[System]`.** `AgentRunRequest.custom_system_prompt` is rejected by the cloud agent (`unknown option '--system-prompt'`); `conversation_state.root_prompt_messages_json` (SHA-256 blob) is fetched by the server but silently ignored by the model. Inlined system text is the only method the account e2e-obeys.
- **Vision via inline `SelectedImage.data` (field 8), not `blob_id`.** Same trap: pure `blob_id` needs a KV blob handshake the gateway does not do; inline bytes sidestep it. Prior art either drops images entirely or substitutes a text placeholder.
- **`GetBlobResult` shape.** `KvClientMessage.get_blob_result` (field 2) is a *message* whose field 1 carries the bytes. Placing raw bytes directly in field 2 makes Cursor read an empty blob and stall on heartbeats until the ~30 s idle timeout. Fixed in `packages/provider-cursor/src/proto/kv.ts`; this is a real latent bug in some prior implementations.
- **Cursor Tab (`StreamCpp`) → Zed edit-prediction.** No cursor gateway we're aware of implements this. The bridge parses Zed's marker-delimited prompt (Zeta 2.1 V0318 / V0615 hashed-regions / FIM / plain), builds a StreamCpp request, and renders the response as `<|marker_1|>...<|user_cursor|>...<|marker_K|><[end▁of▁sentence]>`.
- **Cursor Tab token estimation without the `CountTokens` RPC.** The live `CountTokens` RPC ignores its `model_name` argument and adds ~1 s latency — untenable for a Tab request that must return in tens of ms. Floway uses a corpus-calibrated chars-per-token estimator with a per-model ratio.
- **Cursor Tab client-version impersonation as the Cursor IDE (client-type `ide`, `CURSOR_TAB_CLIENT_VERSION`).** The Tab backend gates on the IDE-style client version; the CLI version string yields empty completions.
- **Observed-context-window caching.** The model catalog fetches the per-model context window from `AvailableModels` tooltip prose (the only pre-observation source), then overlays it with the live observed `ConversationTokenDetails.maxTokens` from the RunSSE stream (cached per `(modelId, maxMode)` with TTL). Prior implementations either hardcode context windows or read them off the tooltip only.

## Per-request token accounting: read this before you look at the graphs

**A turn that ends on a tool call reports zero tokens. Real usage only lands on the last turn of the whole conversation.** This is not a bug — it is the only shape the Cursor upstream gives us — and operators watching per-request usage graphs need to know it up front.

Cursor's cumulative token total is delivered as a `ConversationTokenDetails` checkpoint frame. That checkpoint only arrives at the very end of a run, **after the model's final answer** — i.e., after the last non-tool assistant turn. A turn that pauses at an `exec_request` (any turn that emits a tool call) returns to the gateway before its checkpoint has been written to the stream, so:

- **Tool-call turn** → `usage: { prompt: 0, completion: 0, total: 0 }` — Floway records the request row so the call is still counted, but every dimension is zero.
- **Final answer turn of the conversation** → `usage` carries the entire run's cumulative token totals (`total = usedTokens`, `completion = Σ TokenDeltaUpdate.tokens` clamped to `total`, `prompt = total − completion`).

In per-request usage graphs this reads as: N−1 zero-usage rows followed by one row that carries all the tokens for the whole conversation. Aggregate totals stay correct; per-request medians and percentiles do not, and cannot, without Cursor exposing per-turn totals it currently does not.

Two data-plane changes in this branch back this shape:

- `agent-translate.ts` (finalize path) — computes usage as above and always emits a trailing usage frame, zeros included, so every turn produces a countable request row.
- `packages/gateway/src/data-plane/chat/shared/respond.ts` — `rememberUsage` now keeps the first zero-usage frame instead of dropping it, and `recordUsage` writes a bare request row on a non-null-but-empty usage object. Without both, the N−1 tool-round rows would be silently invisible to the request counter. (The docstring on `SourceStreamState.rememberUsage` explains the invariant for future providers with the same shape.)

Non-agent Cursor traffic (no tools advertised) hits only the final-turn path and looks normal in the graphs.

**Cursor Tab (`/v1/completions`) is separate — and its numbers are estimated, not measured.** The `StreamCpp` backend does not return a usage frame at all, and the live `CountTokens` RPC ignores its `model_name` argument and adds ~1 s latency (untenable for a Tab request that has to return in tens of ms). Floway substitutes a corpus-calibrated chars-per-token estimator with a per-model ratio and records the estimate on the request row. Aggregate Cursor Tab token counts in the graphs are indicative, not authoritative — treat them as sizing signal, not billing truth.

## Upgrading an existing deployment

Anyone who already runs Floway on Cloudflare Workers **must edit their local `wrangler.jsonc`** before `pnpm run deploy` will publish this branch — the deploy gate refuses to publish otherwise. The gate is `scripts/check-wrangler.ts`; when the config drifts it lists every missing entry by path, so the mechanical fix is straightforward, but it does need doing:

- Add a new Durable Object binding `DURABLE_HTTP_SESSION` → class `DurableHttpSessionDO` under `durable_objects.bindings` (alongside the existing `BROADCAST_DO`).
- Append a new DO migration entry to `migrations`: `{ "tag": "v2", "new_sqlite_classes": ["DurableHttpSessionDO"] }` (DO migrations are append-only; do **not** modify the existing `v1`).
- Add `"compatibility_flags": ["nodejs_compat"]` at the top level (required by `@reclaimprotocol/tls/webcrypto`'s `import { webcrypto } from 'crypto'`).
- Verify against the updated `wrangler.example.jsonc` in this PR — the deploy gate now enforces bidirectional mirroring, so any real-config key the example doesn't pin (aside from `account_id`) will also fail.

D1 migrations `0048`–`0050` apply automatically as part of `pnpm run db:migrate:remote`; no manual step there.

## Test plan

- [x] `pnpm run typecheck` — green
- [x] `pnpm run lint` — green
- [x] `pnpm run test` — 4 136 tests, 0 failed
- [x] `wrangler dev --local` boot + `/api/health`, `/v1/models`
- [x] `POST /v1/chat/completions` streaming against a real Cursor account (`claude-sonnet-5`) — end-to-end response received
- [x] `POST /v1/completions` `cursor-tab` plain FIM — real completion received
- [x] `POST /v1/completions` `cursor-tab` Zeta 2.1 (V0318) marker prompt — output rendered as `<|marker_1|>...<|user_cursor|>...<|marker_2|><[end▁of▁sentence]>`
- [ ] `pnpm run db:migrate:remote` on remote D1 (0048, 0049, 0050)
- [ ] `pnpm run deploy` — first-time deploys must add the new `DURABLE_HTTP_SESSION` binding + v2 DO migration + `nodejs_compat` to their `wrangler.jsonc`
- [ ] Post-deploy: prod chat completions call end-to-end
- [ ] Post-deploy: prod Cursor Tab call end-to-end
- [ ] Post-deploy: cross-request tool-result continuation (validates DO acquire/release + D1 session single-flight + outbound-socket survival on a real DO)
- [ ] Post-deploy: idle-alarm eviction (session idle > 5 min → DO tears down before the 15-min runtime cap)


packages/gateway/migrations/0050_cursor_drop_quotasnapshot.sql
```diff
@@ -0,0 +1,20 @@
...
+-- without the key are left unchanged. SQLite's json_remove treats a missing
+-- path as a no-op.
+
+UPDATE upstreams
+SET state_json = json_replace(
+  state_json,
+  '$.accounts[0]',
+  json_remove(json_extract(state_json, '$.accounts[0]'), '$.quotaSnapshot')
+)
+WHERE provider = 'cursor'
```
Menci, line 18: You should NEVER UPDATE a freshly CREATE-d table (with ABSOLUTELY NO DATA) immediately...

packages/gateway/src/data-plane/shared/passthrough-attempt.ts
```diff
@@ -56,6 +56,7 @@ export const passthroughAttempt = async (args: PassthroughAttemptArgs): Promise<
     recordUpstreamLatency: recorder.record,
     waitUntil: ctx.backgroundScheduler,
     headers: inboundHeadersForUpstream(c),
+    apiKeyId: ctx.apiKeyId,
```
Menci, line 59: Land these prequirements changes separately before this PR

PR #160, opened 2026-07-09
feat(api-keys): switch to sk-format tokens with an optional caller-provided key

chiyuki0325:
original 64-characters API key is too long to type manually and hard to identify
switch to OpenAI-format sk-(48 chars) api key format for better distinguishability.
also added custom API key feature.


chiyuki0325:
new key format: `^sk-[A-Za-z0-9]{20}T3BlbkFJ[A-Za-z0-9]{20}$`

packages/gateway/migrations/0050_api_key_format.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE api_keys ADD COLUMN api_key_format TEXT NOT NULL DEFAULT 'openai';
```
Menci: Why need this?

Why database needs to know "format"?

chiyuki0325: The format is used during rotation to generate a new API key in the corresponding format. For "custom" format, user input will be requested

PR #164, opened 2026-07-10
ci(build): tag non-main refs with sanitized branch name

yyyr-p:
## Summary

- Non-`main` refs now also emit `<image>:<sanitized-branch>` alongside `:latest` and `:<date>.<run_id>`, so downstream stacks (e.g. a `docker-compose.*.yml` on a fork) can pin to a stable per-branch image without needing to track each run's ephemeral date/id.
- Slashes in branch names (`fix/foo`) become `-` because docker tags disallow `/`.
- `main` still produces only `:latest` and `:<date>.<run_id>` — no redundant `:main` tag.

Stacked on #163 — that PR adds the `OWNER` env-plumbing pattern this one reuses (adds `REF_NAME`) and drops the top-level `IMAGE_PREFIX` env; landing this on its own would conflict with the same block.

## Test plan

- [ ] Dispatch on a non-`main` ref and confirm the manifest for `:latest`, `:<date>.<run_id>`, and `:<branch>` are all the same digest.
- [ ] Dispatch on `main` and confirm no `:main` tag is produced.
- [ ] Dispatch on a branch containing `/` (e.g. `fix/foo`) and confirm the emitted tag is `:fix-foo` (not rejected by the registry).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,9 +7,6 @@ on:
   schedule:
     - cron: '0 6 * * *'
 
-env:
-  IMAGE_PREFIX: menci/floway
-
 jobs:
   build:
     permissions:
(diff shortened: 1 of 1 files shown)


Menci:
Closed as duplicatation of #163. I see the change is "derive image name from repo owner".

PR #167, opened 2026-07-11
fix(web): keep Settings usable when the model catalog fails to load

Dustella:
## Summary

- load the Settings page's required data concurrently, then fetch the raw model catalog
- keep the route in its navigation-loading state until the catalog request settles, so Settings renders once with complete content instead of filling model-derived fields after mount
- surface catalog failures inline while keeping Settings usable after the failed request completes
- deduplicate concurrent model-store loads so repeated callers share one HTTP request
- cover the empty-gateway contract: `/api/models?aliases=false&include_unlisted=true` returns an empty 200 response when there are truly no upstreams

## Original reproduction

1. Run the Node gateway on `http://localhost:8788` and the Vite dashboard on `http://localhost:5174`.
2. Import a ChatGPT/Codex subscription credential without configuring a working network path or proxy.
3. Navigate to **Settings**.
4. The route waits for live model discovery; repeated attempts previously created overlapping `/api/models` requests.

During diagnosis, the local database contained an enabled `codex` upstream with an empty proxy fallback list and no model-cache row. No credential material is included in this PR. A database with genuinely zero upstream rows already follows a different, valid path and returns an empty catalog.

## Root cause

The Settings loader originally fetched all page data and the live model catalog in the same `Promise.all`, allowing catalog discovery to race data needed to construct the upstream and alias views. Moving catalog loading into component setup avoided that race but introduced a third visual state: Settings rendered first and model-derived counts and labels filled in later.

The module-scoped models store also shared state without sharing an in-flight Promise. Concurrent callers could therefore start duplicate catalog requests.

## Fix

The Settings data loader now completes its required page-data `Promise.all` first, then awaits `useRawModelsStore().load()`. This preserves the required ordering while keeping navigation in its loading state until all visible Settings content is ready. A catalog failure is retained in the store and rendered as an inline error once the page opens.

Each models-store variant owns one in-flight Promise. Concurrent `load()` calls join that Promise, and the entry is cleared after success or failure so a later explicit reload can retry.

## Expected behavior

- Required Settings data loads concurrently; the model catalog request starts afterward.
- Settings does not render a partially populated intermediate state.
- Concurrent model-store loads do not create duplicate catalog requests.
- A catalog 502 appears as an inline error after the request settles.
- Upstream, proxy, alias, search, import, and export controls remain available after a catalog failure.
- A gateway with no upstream rows receives an empty 200 catalog.

## Verification

- `pnpm --filter @floway-dev/web run test` — 17 files, 152 tests passed
- `pnpm --filter @floway-dev/web run typecheck` — passed
- `git diff --check` — passed

### DIFF
--- a/apps/web/src/pages/dashboard/settings.vue
+++ b/apps/web/src/pages/dashboard/settings.vue
@@ -32,7 +32,6 @@ export const useSettingsPageData = defineBasicLoader(async () => {
   const [searchRes] = await Promise.all([
     callApi<SearchConfig>(() => api.api['search-config'].$get()),
     useUpstreamsStore().load(),
-    useRawModelsStore().load(),
     useProxiesStore().load(),
     useModelAliases().load(),
     useRuntimeInfo().load(),
@@ -51,6 +50,10 @@ definePage({ meta: { requiresAdmin: true } });
 const router = useRouter();
 const { upstreams, loading: storeLoading, load } = useUpstreamsStore();
 const modelsStore = useRawModelsStore();
+// The model catalog enriches counts and alias target labels, but it is not a
+// prerequisite for editing upstreams or proxies. Start it after navigation
+// so an unreachable upstream cannot hold the entire Settings route open.
+void modelsStore.load();
 const proxiesStore = useProxiesStore();
 const aliasesStore = useModelAliases();
 const { load: loadProxies } = proxiesStore;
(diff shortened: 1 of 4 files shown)


apps/web/src/pages/dashboard/settings.vue
```diff
@@ -32,7 +32,6 @@ export const useSettingsPageData = defineBasicLoader(async () => {
   const [searchRes] = await Promise.all([
     callApi<SearchConfig>(() => api.api['search-config'].$get()),
     useUpstreamsStore().load(),
-    useRawModelsStore().load(),
```
Menci, line 35: This makes the page load flickering. A page navigation should NOT have a third state than "NO page content" and "ALL page content"

Menci (changes requested):

apps/web/src/pages/dashboard/settings.vue
Dustella: resolved in 834a41e524a2259d1d5f21bac66f6e11eadee9b0, moving `modelsStore.load();` to `useSettingsPageData()` as expected

Dustella: the fix still works, the page won't be dead, but will load 10s if upstream is unreachable.

PR #172, opened 2026-07-11
fix(responses): preserve hosted tool semantics through the shim

Menci:
## What

The Responses server-tool shim replaces hosted declarations with one internal function before upstream inference, then restores the hosted declaration in synthesized response envelopes. Duplicate hosted declarations therefore need one consistent selection rule across request execution and response echo.

Live bidirectional probes confirmed the upstream behavior:

- Azure and Copilot accept duplicate hosted `web_search` declarations, deduplicate them to one tool, and retain the last complete declaration.
- Azure treats `web_search` and `web_search_preview` as one family, preserving the last alias, filters, location, and context size.
- Azure applies the same last-entry rule to duplicate hosted `image_generation` declarations.

Update the shared hosted-tool boundary to validate every matching declaration, select the last complete declaration, inject one collision-resolved function at the first declaration's position, and restore that selected declaration in downstream echoes. Web-search runtime filters now come from the same last declaration, so backend execution and response metadata cannot disagree.

Capture a forced hosted `tool_choice` before request rewriting. Multi-turn server-tool execution demotes its internal function choice to `auto` after the first turn; synthesized responses now restore the original hosted choice even when the terminal upstream envelope omits `tool_choice`.

Namespaces remain client-owned function/custom tool groups. This PR does not consume, strip, translate, or execute Codex namespaces.

Sources:

- https://github.com/openai/openai-node/blob/1cdc0196b4341ee641ec6839e08744b2771250e4/src/resources/responses/responses.ts#L937-L976
- https://github.com/openai/openai-node/blob/1cdc0196b4341ee641ec6839e08744b2771250e4/src/resources/responses/responses.ts#L3346-L3458

## Test plan

- [x] Reversed hosted web-search declarations select the last alias and complete configuration
- [x] Web-search backend filters match the selected hosted declaration
- [x] Duplicate hosted image-generation declarations retain last-entry configuration
- [x] A forced hosted image choice survives multi-turn demotion and an omitted terminal echo
- [x] Focused hosted-tool suites — 369 tests pass
- [x] Full workspace tests — 350 files / 4308 tests pass
- [x] Full workspace typecheck and lint


Menci:
Claude Code agent update: production Azure `gpt-5.4` + `gpt-image-2` probing confirmed the broader shim behavior under review.

- Azure requires `x-ms-oai-image-generation-deployment: gpt-image-2` for native Responses `image_generation`.
- A client `namespace: image_gen` is rejected as a reserved namespace, alone or alongside hosted `image_generation`, regardless of order.
- Duplicate hosted `image_generation` declarations are accepted and deduplicated; the last declaration's config wins.

I am extending this PR so the gateway image shim consumes the Codex namespace before the Copilot fallback strip, canonicalizes alternate image declarations to one replacement tool, preserves/restores the original namespace/choice shape downstream, and follows deterministic last-declaration precedence. The existing Copilot strip remains the disabled-shim safety net.

Menci:
Codex agent probe update: duplicate server-tool precedence has now been tested with distinguishable declarations instead of identical controls.

Azure OpenAI `gpt-5.4` and GitHub Copilot `gpt-5.4` produced the same result for Responses `web_search`:

- `[search_context_size: low, search_context_size: high]` echoed one tool with `high`
- the reversed order echoed one tool with `low`
- Azure also deduplicated `web_search` / `web_search_preview` as one family and preserved the last alias, filters, and configuration

Every request used `tool_choice: "none"`, `stream: false`, and `max_output_tokens: 16`; no search was executed. Azure was called directly using the configured production upstream. Copilot was also called directly because the production upstream's `ss://` proxy cannot be used by curl, so the Copilot egress did not match production; token exchange, `/models`, and all four `/responses` controls succeeded.

The earlier Copilot probe used two identical `{ "type": "web_search" }` entries, so it proved deduplication but could not distinguish first-wins from last-wins. The new bidirectional controls show that both upstreams use last-declaration precedence, matching the existing Azure duplicate `image_generation` result. The PR will therefore remove `declarationPrecedence` and use one dedupe-to-last rule for recognized server-tool declarations.

Menci:
Codex agent correction: the earlier plan to consume Codex's `image_gen.imagegen` namespace in the gateway was based on an incorrect execution-ownership assumption and has been removed from this PR.

OpenAI's Responses contract uses namespaces to group function/custom tools; a namespaced function call carries `namespace` and `name`, and its output is supplied by the API consumer. Current Codex registers `image_gen.imagegen` as a local `ToolExecutor`, executes the provider-relative Images API, saves the artifact, and returns `function_call_output`. Treating that namespace as an alternate hosted `image_generation` declaration would replace the Codex executor and lose its client-side behavior.

Sources:

- https://github.com/openai/openai-node/blob/1cdc0196b4341ee641ec6839e08744b2771250e4/src/resources/responses/responses.ts#L937-L976
- https://github.com/openai/openai-node/blob/1cdc0196b4341ee641ec6839e08744b2771250e4/src/resources/responses/responses.ts#L3346-L3458
- https://github.com/openai/codex/blob/8604689ec5e3437eb79802d8d72249b7722fbf5b/codex-rs/ext/image-generation/src/tool.rs#L89-L218

The final PR now changes hosted tools only: repeated hosted `web_search` / aliases and `image_generation` declarations use last-entry precedence, runtime web-search configuration matches the selected declaration, and forced hosted choices survive the shim's multi-turn demotion. Namespace protocol handling, provider stripping, and execution remain unchanged from `main`.

PR #174, opened 2026-07-11
feat(responses): expose reasoning.context as an open string

Menci:
## What

Add the official OpenAI Responses `reasoning.context` field to Floway's protocol type as an open string:

- documented values: `auto`, `current_turn`, `all_turns`
- unknown future values remain valid and pass through verbatim
- `null` matches the official SDK contract

Runtime payload preservation already worked through Floway's object-spread boundaries; this PR makes that contract explicit and locks it in with tests.

Cross-protocol behavior is intentionally asymmetric:

- native Responses payloads preserve `context`, including future values and alias effort/summary overlays
- Messages, Chat Completions, and Gemini do not have an equivalent source field, so their `*-via-responses` translators omit rather than invent it
- reverse target protocols have no destination slot and continue dropping it

This PR does not enable Responses Lite or force encrypted reasoning content.

Sources:

- https://github.com/openai/openai-python/blob/f16fbbd2bd25dc1ff150b5f78dbd15ff6bab6d91/src/openai/types/shared/reasoning.py#L19-L25
- https://github.com/openai/openai-node/blob/61539248cbe04665de68a71e6fd878127ae4db87/src/resources/shared.ts#L262-L269

## Test plan

- [x] `pnpm --filter @floway-dev/protocols run typecheck`
- [x] `pnpm --filter @floway-dev/translate run typecheck`
- [x] ESLint on all changed TypeScript files
- [x] Protocols tests — 22 files / 110 tests
- [x] Translate tests — 26 files / 454 tests
- [x] Gateway tests — 145 files / 1792 tests
- [x] Native known/future context values preserve verbatim
- [x] Messages, Chat Completions, and Gemini translators never synthesize context
- [x] Alias effort/summary overlays preserve an existing context

### DIFF
--- a/packages/gateway/src/data-plane/model-aliases/apply-rules_test.ts
+++ b/packages/gateway/src/data-plane/model-aliases/apply-rules_test.ts
@@ -111,15 +111,18 @@ test('responses: budget_tokens / adaptive have no native slot — silently dropp
   assertEquals('adaptive_thinking' in body, false);
 });
 
-test('responses: alias rules overwrite existing reasoning + service_tier fields', () => {
-  const body = resPayload({ reasoning: { effort: 'low', summary: 'auto' }, service_tier: 'default', text: { verbosity: 'high' } });
+test('responses: alias rules overwrite owned fields while preserving reasoning.context', () => {
+  const body = resPayload({
+    reasoning: { effort: 'low', summary: 'auto', context: 'future_mode' },
+    service_tier: 'default',
+    text: { verbosity: 'high' },
+  });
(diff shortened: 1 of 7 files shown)


Menci:
Claude Code agent update: all `*-via-responses` paths have been audited. Chat Completions and Gemini have no source field equivalent to Responses `reasoning.context`, so they must omit rather than synthesize it; reverse targets likewise have no slot. The branch now adds explicit no-invention tests for Chat→Responses and Gemini→Responses, plus a test that a native Responses context survives effort/summary alias overlays. No additional runtime mapping is appropriate.

PR #176, opened 2026-07-11
feat(pricing): account for Responses cache writes and GPT-5.6 input tiers

Menci:
## What

Add standard OpenAI Responses cache-write accounting and the GPT-5.6 input-length pricing axis.

- model `usage.input_tokens_details.cache_write_tokens` as an optional-compatible standard field
- split Responses usage into disjoint ordinary input, cache read, and cache write buckets
- map cache writes to Messages `cache_creation_input_tokens` in streaming and non-streaming translations
- add `ModelPricing.inputLengthTiers`, orthogonal to the existing wire `service_tier` overlays
- select the input tier per request before usage persistence
- persist `input_tier` in usage buckets through migration `0051`
- add Copilot and Codex prices for GPT-5.6 Sol/Terra/Luna, including the full-request `>272k` tier

The resolver does not invent an undocumented priority × long-context multiplier. A long-context tier may publish an explicit service-tier combination later; until then, its complete published rates win.

## Pricing

USD per million tokens, input/read/write/output:

- Sol: `5 / .5 / 6.25 / 30`; above 272k: `10 / 1 / 12.5 / 45`
- Terra: `2.5 / .25 / 3.125 / 15`; above 272k: `5 / .5 / 6.25 / 22.5`
- Luna: `1 / .1 / 1.25 / 6`; above 272k: `2 / .2 / 2.5 / 9`

Sources include the generated OpenAI Python/Node SDK field definitions and the OpenAI Codex GPT-5.6 migration guide; caozhiyuan/copilot-api@5a28eee is used only as a cross-check.

## Migration

`0051` recreates `usage` and `usage_requests` so `input_tier` participates in the bucket uniqueness key. Existing rows backfill `NULL` and retain their prior pricing behavior.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm run test` — 335 files / 4011 tests pass
- [x] SQL usage-repo tests apply all migrations, including `0051`, against a real sql.js database
- [x] Tests cover cache-write extraction/translation, malformed bucket rejection, input-tier selection and service-tier composition, provider pricing, aggregation, and import/export


Menci:
Claude Code agent here. Closing this implementation while I rework the pricing model from the review: input-length and service-tier pricing need an explicit Cartesian-product representation, and I am re-verifying whether the 272k rule reprices the whole request or only the excess. I will open a replacement PR rather than patch this design in place.

PR #177, opened 2026-07-11
feat(codex): restore first-frame tunneled errors to HTTP

Menci:
## What

Restore Codex Responses errors that arrive as the first SSE frame with tunneled HTTP metadata back to a normal HTTP error before the downstream stream is committed.

At the provider-codex `/responses` boundary:

- peek exactly one parsed frame
- empty stream → HTTP 502 provider failure
- first generic `{ type: "error", status/status_code, headers, error }` frame → provider failure using the valid 4xx/5xx status, otherwise 502
- merge safe upstream/event headers and preserve the error body
- replay every normal first frame exactly once
- leave `response.failed`, `response.incomplete`, later errors, Copilot, Chat, and Messages unchanged

Returning `ok:false` feeds Floway's existing candidate fallback. Iterator cancellation and abort propagation are preserved.

This one-frame lookahead delays downstream response headers until the first upstream event. It does not change TTFT's output-token definition; error/lifecycle frames are not output tokens.

Evidence:

- https://github.com/caozhiyuan/copilot-api/commit/141f86fa0376f7fd87216a80e621ecc4a2e7325b
- https://github.com/caozhiyuan/copilot-api/commit/0034feb928404fffd28de937561b199f830e89fd
- current openai/codex distinguishes standard `response.failed` lifecycle events from generic tunneled errors

## Test plan

- [x] `pnpm run test` — 333 files / 4007 tests pass
- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] 14 cases cover status aliases/fallback, body and header preservation, empty stream, replay, standard lifecycle exclusions, later errors, cancellation, and iterator failure

### DIFF
--- a/TRANSLATION.md
+++ b/TRANSLATION.md
@@ -196,6 +196,19 @@ inline and rebuilds the envelope client-side).
   `(instructions + first user-message text)` so the upstream prompt cache
   hits across turns of the same conversation (~88% input-token cache hit
   measured against gpt-5.4)
+- promotes a tunneled HTTP error back to a real failure. The Codex backend
+  answers `200 text/event-stream` and, when the call fails after the stream
+  is open, emits a first-frame `{ type: "error", status, headers, error }`
+  event instead of `response.failed`. The provider peeks exactly the first
+  parsed frame: an empty stream becomes `502`; a first-frame tunneled error
+  becomes an `ok:false` HTTP response (status from `status`/`status_code`
+  when a valid `400..599`, else `502`; body preserving the upstream error;
+  headers merged from the upstream response and the event, minus hop-by-hop
+  and body-framing), which flows into the normal candidate fallback. A
+  normal first frame is replayed unchanged. Only the first frame is
+  eligible — a later `error` is mid-stream failure owned by the client. The
+  one-frame lookahead delays downstream TTFB to the first event; the TTFT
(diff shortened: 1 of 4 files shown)


Menci:
Claude Code agent here. Closing this after re-checking the evidence against the provider boundary. The caozhiyuan change was gated by the downstream Codex User-Agent in a generic provider Responses handler; that establishes a Codex-client compatibility need, not that OpenAI's official Codex backend emits malformed tunneled HTTP errors. Official openai/codex treats / as valid SSE lifecycle failures, while our authenticated live Codex probe could not reproduce a generic first-frame error. Putting this workaround in provider-codex was therefore an unsupported attribution. I will not replace it unless we capture the official backend emitting the generic tunneled shape.

Menci:
Claude Code agent correction: the prior close comment's inline code was consumed by the shell. The intended evidence is that official openai/codex treats `response.failed` and `response.incomplete` as valid SSE lifecycle failures. The caozhiyuan change was originally gated by the downstream Codex User-Agent in a generic provider handler; that does not establish that OpenAI's official Codex backend emits malformed generic first-frame errors. Our authenticated live probe could not reproduce one. Provider-codex was therefore the wrong boundary, so this PR remains closed.

PR #178, opened 2026-07-11
feat(pricing): support explicit multidimensional rates

Menci:
## Summary

Replace legacy model cost overlays and implicit per-dimension fallbacks with explicit multidimensional pricing entries.

- Represent each rate card as `model.pricing.entries`, with one complete rates vector and an optional selector.
- Require exactly one Base entry and identical rate dimensions across every entry.
- Support open-string `serviceTier` equality coordinates and whole-request `inputTokens` `gt` / `gte` thresholds through one typed axis registry.
- Select the highest matching global or equality-scoped threshold, then perform one exact selector lookup. Missing selector combinations use the whole Base vector; rates are never merged field-by-field.
- Collect structured pricing issues once in `@floway-dev/protocols/common` and reuse their entry indexes and dimension differences in backend validation and dashboard errors.

## Persistence and Transfer

- `0053_usage_pricing_selector.sql` replaces the service-tier bucket column with canonical JSON `pricing_selector`.
- `0054_model_pricing.sql` materializes legacy overlays and fallback semantics into uniform explicit vectors while preserving historical realized cost.
- Usage rows store the selected canonical coordinate plus an immutable resolved `rates` snapshot; later catalog changes cannot rewrite historical usage.
- Data-transfer format v9 exports `pricingSelector` and `rates`, requires exact token/rate maps, and rejects historical formats.
- Embedded provider catalog changes bump `MODEL_CATALOG_REVISION` from 1 to 2.

## Dashboard Authoring

The model editor uses one focused `PricingEditor` inside the existing card layer.

- Wide layouts use proportional navigation/editor tracks; narrow containers stack navigation above the editor.
- Entry controls stay in navigation while selectors and rates stay in the editor.
- Combined coordinates render as `priority, > 512000 tokens`; the compact threshold button toggles `>` / `>=`.
- Adding an entry clones Base rates and preserves every dimension present in any draft.
- All validation errors are grouped below the form and identify affected entry numbers.
- Every manual model is validated through the canonical backend parser, including unselected rows and JSON editing mode.

Malformed custom-upstream pricing remains best-effort catalog metadata: Floway drops the complete pricing block without dropping the model.

## Risk

The two migrations rewrite persisted usage bucket identity and manual model pricing JSON. Export v9 intentionally rejects prior formats. Catalog revision 2 makes pricing-bearing model snapshots cold once after deployment.

## Test Plan

- [x] `pnpm run test`
- [x] `pnpm --filter @floway-dev/provider-ollama run test`
- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm --filter @floway-dev/web run build`
- [ ] Dashboard pricing editor smoke test: navigation, threshold toggle, grouped numbered validation, and responsive stacking


Menci:
Claude Code agent update: this PR is paused for a broader architecture comparison. The strict `>272k` whole-request math and explicit priority×context cells are verified, but I am now comparing how LiteLLM, models.dev, Portkey, Helicone, OpenRouter, Cloudflare and other gateways represent context-length pricing alongside service tiers. I will not treat the current `inputLengthTiers`/`input_above_tokens` design as final until that multi-implementation review is complete.

Menci:
Claude Code agent architecture result after comparing LiteLLM, models.dev, Portkey, Helicone, Langfuse, OpenRouter, and Cloudflare's public contract:

- context length is a pricing selector independent of wire `service_tier`; it is not another token billing dimension and not a subtype of service tier
- crossed thresholds select a whole-request price, not marginal overage
- selector combinations are often sparse explicit cells; missing cells must not be auto-multiplied or silently inherit an unrelated axis
- persistence should retain `tier` and `input_above_tokens` as separate bucket coordinates

The current nested `inputLengthTiers[].tiers` shape still privileges one axis. I am refactoring this PR to a symmetric flat list of explicit selector coordinates and complete price vectors. Empty selector is the base cell; exact missing combinations remain unpriced. The existing migration coordinates and cache-write accounting stay.

The review also found that the new coordinate changes backup identity, so the export/import format will be bumped to version 9 rather than silently extending version 8.

PR #183, opened 2026-07-12
feat: passthrough OpenAI tool_search family; flatten-tool-search-family flag

yyyr-p:
## Summary

Floway 502'd on any `/v1/responses` request whose `input[]` carried a `type: 'additional_tools'` item — the gpt-5.4+ OpenAI Responses API tool_search feature family (`additional_tools` input items, `type: 'namespace'` tool containers, `type: 'tool_search'` / `programmatic_tool_calling` hosted tools, `defer_loading` / `allowed_callers` function-tool fields, `program` / `program_output` output items) was entirely unknown to the protocol / store / translate layers. `stageInputItem` threw a TypeError for the unknown item type; further latent gaps existed for the family's other members that would surface as clients start using them.

The change lands in two layers:

**Layer 1 — unconditional passthrough (no flag):** make Floway an opaque gateway for the whole family, per the repo's `Upstream Models And Field Values Are Opaque` design principle. Protocol union collects the new input/output item variants, hosted-tool discriminator, function/custom tool fields, and `function_call.caller`. Store registers `program: 'pg'` / `program_output: 'pgo'` prefixes so PTC output items survive snapshot storage, and skip-persists `additional_tools` (turn-scoped developer directive re-declared verbatim by the client on every request — mirrors `compaction_trigger`).

**Layer 2 — opt-in flatten to legacy shape (`flatten-tool-search-family` flag):** for upstreams that do not implement the family (translated Chat Completions / Messages targets; Responses endpoints stuck below gpt-5.4), one interceptor rewrites the request in a single pass — move ` additional_tools[].tools` into top-level ` payload.tools[]`; unpack `namespace` containers (sub-tool names prefixed \` <namespace>__ \`); drop `tool_search` / `programmatic_tool_calling` entries; strip `defer_loading` / `allowed_callers`. A response-side interceptor scans the downstream event stream and reverses the name-prefixing on `function_call` (bare name) and `custom_tool_call` (bare name + `namespace` field) output items so clients match against their originally-declared tool registry.

Provider defaults: `false` for OpenAI-native (codex / copilot / azure / custom); `true` for translated targets (ollama / claude-code) that need the desugar to make the request survive translation.

Also fixes a pre-existing silent-drop bug in \`responses-via-chat-completions\` and \`responses-via-messages\` translators — their \`translateResponsesTools\` narrowed to \`function\` / \`custom\` and let \`namespace\` container tools fall through, silently losing the nested sub-tools. Both now apply the same shared \`flattenToolSearchFamilyTools\` helper unconditionally so sub-tools survive translation with the same \`<namespace>__\` prefix that the response-side interceptor reverses.

Structured as 6 commits following the layering:

1. `fix(responses): accept OpenAI tool_search family artifacts on passthrough`
2. `feat(protocols): tool_search family desugar helpers`
3. `fix(translate): unpack namespace sub-tools instead of silently dropping`
4. `feat(flags): flatten-tool-search-family for legacy upstreams`
5. `feat(responses): flatten-tool-search-family request/response interceptors`
6. `test(responses): add fixture-based tool_search family e2e integration test`

Refs:
- <https://developers.openai.com/api/docs/guides/tools-tool-search>
- <https://developers.openai.com/api/docs/guides/tools-programmatic-tool-calling>

## Test plan

- [x] Unit: \`unpackNamespaceTools\` / \`unprefixNamespaceToolCall\` / \`flattenToolSearchFamilyTools\` helpers — edge cases (empty namespace, malformed \`tools\`, multi-\`__\` delimiter, all six flatten targets present).
- [x] Unit: \`withFlattenToolSearchFamily\` interceptor — flag on runs the six-op desugar; flag off is a no-op; no-op detection preserves reference identity.
- [x] Unit: \`withUnprefixNamespaceToolCalls\` interceptor — \`function_call\` bare-name reverse; \`custom_tool_call\` bare-name + \`namespace\` field populate; \`response.completed\` envelope rewrite; passthrough of non-\`__\` names; flag off is a no-op.
- [x] Regression: \`additional_tools\` and \`compaction_trigger\` throwing in \`prefixForItemType\` (skip-persist invariant pins).
- [x] Integration: fixture-based e2e test drives \`responsesAttempt.generate\` against a mocked chat-completions candidate with the full tool_search family payload — asserts (a) outbound Chat Completions body captured is fully desugared; (b) upstream tool_call arrives at the client with a bare name; (c) flag off causes the translator to reject \`additional_tools\` (proving the fold interceptor is what removes it on flag on).
- [x] Workspace: \`pnpm -r run typecheck\`, \`pnpm run lint\`, \`pnpm run test\` (4016 tests) all green at HEAD.

### DIFF
--- a/packages/gateway/src/data-plane/chat/responses/interceptors/flatten-tool-search-family.ts
+++ b/packages/gateway/src/data-plane/chat/responses/interceptors/flatten-tool-search-family.ts
@@ -1,0 +1,72 @@
+// Desugar the gpt-5.4+ Responses API tool_search feature family into a legacy
+// tools[]-only shape before dispatch. Always-attached; flag-gated by
+// `flatten-tool-search-family`.
+//
+// The family features are correlated (all released together as one bundle and
+// depend on each other): implementing this flag as a single one-pass rewrite
+// matches every real upstream population — either the endpoint supports the
+// whole family (leave verbatim, flag off) or it supports none of it (desugar
+// the entire family in one place, flag on).
+//
+// Outbound (request → upstream):
+//
+// 1. Every `type: 'additional_tools'` input item is removed; its inner
+//    `tools[]` is appended to the top-level `payload.tools[]`.
+// 2. Every `type: 'namespace'` container in the resulting `payload.tools[]`
+//    is unpacked into its nested sub-tools (via `unpackNamespaceTools`,
+//    which also prefixes each sub-tool's `name` with `<namespace>__` to
+//    preserve grouping semantic — see `withUnprefixNamespaceToolCalls` for
+//    the response-side inverse).
+// 3. Every remaining `type: 'tool_search'` or `type: 'programmatic_tool_calling'`
+//    hosted-tool entry is dropped from `payload.tools[]`.
+// 4. Every remaining `function`/`custom` tool has its `defer_loading` and
+//    `allowed_callers` fields stripped (they have no meaning without
+//    `tool_search` / PTC).
+//
+// Inbound: nothing — see `withUnprefixNamespaceToolCalls` for the response-side
+// pair that undoes the namespace-name prefix on tool-call outputs.
+//
+// Ref: https://developers.openai.com/api/docs/guides/tools-tool-search
+
+import type { ResponsesInterceptor } from './types.ts';
+import { flattenToolSearchFamilyTools, type ResponsesInputItem, type ResponsesTool } from '@floway-dev/protocols/responses';
+import { providerModelOf } from '@floway-dev/provider';
+
+export const withFlattenToolSearchFamily: ResponsesInterceptor = async (ctx, _request, run) => {
+  if (!providerModelOf(ctx.candidate).enabledFlags.has('flatten-tool-search-family')) return await run();
+
+  const originalInput = ctx.payload.input;
+  const extracted: ResponsesTool[] = [];
+  const remainingInput: ResponsesInputItem[] = [];
+  for (const item of originalInput) {
+    if (item.type === 'additional_tools') {
+      const inner = (item as { tools?: unknown }).tools;
+      if (Array.isArray(inner)) extracted.push(...(inner as ResponsesTool[]));
+      continue;
+    }
+    remainingInput.push(item);
+  }
+
+  const originalTools = ctx.payload.tools ?? null;
+  const merged = extracted.length > 0
+    ? [...(originalTools ?? []), ...extracted]
+    : (originalTools ?? []);
+  const flat = merged.length > 0 ? flattenToolSearchFamilyTools(merged) : [];
+
+  const inputChanged = remainingInput.length !== originalInput.length;
+  // Reference-equality on the array + members catches "identical shape" — if
+  // no member differs and the length matches the original tools list, the
+  // rewrite would be a no-op.
+  const toolsChanged =
+    (originalTools === null && flat.length > 0)
+    || (originalTools !== null && (flat.length !== originalTools.length || flat.some((t, i) => t !== originalTools[i])));
+
+  if (!inputChanged && !toolsChanged) return await run();
+
+  ctx.payload = {
+    ...ctx.payload,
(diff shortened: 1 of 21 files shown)


yyyr-p:
## Known limitations of the flatten-tool-search-family interceptors

The `flatten-tool-search-family` flag makes the tool_search family survive routing to a non-native-Responses endpoint (translated Chat Completions / Messages targets, or legacy pre-gpt-5.4 OpenAI-compat proxies), but it does so by desugaring to a legacy shape — some family semantics have no equivalent on the target side and are unavoidably lost. Two are worth flagging as follow-up work rather than accepted tradeoffs:

### 1. `programmatic_tool_calling` — currently only a workaround

The interceptor drops the `type: 'programmatic_tool_calling'` entry from `payload.tools[]` and strips the `allowed_callers: ['programmatic']` restriction from function tools so they become model-callable directly. The model then no longer has access to the JS runtime that would let it batch/orchestrate multiple tool calls in one turn — every tool call becomes a separate model-driven round trip.

This is a **capability regression**, not a lossless translation:

- The model can still call each individual tool, but only one at a time via normal function-call turns.
- Any tool whose only intended caller was the PTC program (`allowed_callers: ['programmatic']`) becomes unrestricted; a client that declared such a tool assuming PTC-only access loses the constraint entirely.
- Response items `program` / `program_output` / `function_call.caller.type: 'program'` are Layer-1-accepted on the ingress side (so passthrough on native paths works) but the model doesn't emit them on the desugared path because PTC isn't in its toolbox anymore.

A **proper fix** would require Floway to synthesize a PTC executor: expose a shim tool the model can call with JS code, run that code in a sandboxed V8 isolate (or equivalent), route its inner tool invocations through the same interceptor chain, and aggregate the results back into a `program_output`-shaped envelope. Non-trivial infrastructure. Should be tracked for a follow-up; the current drop is documented in the flag description as "loses PTC orchestration" so operators know what they're opting into.

### 2. `additional_tools` — moving to top-level `tools[]` breaks prompt caching

Native semantics: `additional_tools` is a **turn-scoped input item** that sits inline in the conversation at the position it was declared. Upstreams that support it can cache the request prefix up to that point unchanged across turns; only content at or after the item participates in re-hashing.

Desugared semantics (flag on): the interceptor moves the item's `tools[]` into top-level `payload.tools[]` and drops the item from `input[]`. From the upstream's cache-key perspective, the `tools[]` field is now serialized at the top of the request and any change to the client's `additional_tools` set (adding a tool, changing a description, renaming a sub-tool) rewrites `tools[]` — which invalidates the entire prefix cache for every subsequent turn.

Impact: repeat conversations that use `additional_tools` may see substantially degraded prompt-cache hit rates on the desugared path — potentially the difference between cache-hit pricing and full-input pricing on every turn after the first tool-list change.

There is **no workaround inside this PR**: target protocols (Chat Completions, Messages) have no equivalent of a mid-conversation tool-declaration item, so preserving cache-friendly positioning is structurally impossible for the family desugar. Operators aware of this can mitigate by keeping `additional_tools` sets stable across turns (or accepting the cost).

### Minor additional losses

For completeness, the flag also drops:

- **`defer_loading: true`** — all sub-tool schemas are declared eagerly; the token savings that `tool_search`-driven lazy loading would deliver are gone.
- **`custom_tool_call.namespace` round-trip fidelity for function-typed sub-tools** — `function_call` has no `namespace` field in the Responses schema, so the response side can only strip the `<namespace>__` prefix back to bare name; namespace context is lost. (`custom_tool_call` sub-tools do keep the namespace, because that item type has the field.)

These two are unlikely to warrant a follow-up on their own but are surfaced here so the tradeoff surface of the flag is fully documented.

### Where these are documented in the codebase

- Flag description in `packages/provider/src/flags.ts` ends with "Loses `defer_loading` eager-vs-lazy semantics, PTC orchestration, and `custom_tool_call.namespace` round-trip fidelity." — surfaces losses to operators.
- Interceptor file headers point at the same tradeoffs.

The above should be tracked as follow-up issues; happy to open them separately if useful.

PR #226, opened 2026-07-16
fix(gateway): add X-Accel-Buffering: no to every SSE response

yyyr-p:
## Summary

Every SSE response Floway emits now carries `X-Accel-Buffering: no`, so streams flow through nginx (and other nginx-based reverse proxies) without `proxy_buffering` holding chunks in the edge until the buffer fills. The header is a no-op on Cloudflare Workers, so the Workers deployment is unaffected; the Node deployment target behind nginx is where this fixes long silent gaps in Anthropic Messages / OpenAI Responses streams.

## What changed

- **New `packages/gateway/src/streaming/sse.ts`** — a thin wrapper around Hono's `streamSSE` that sets `X-Accel-Buffering: no` on the response header before delegating. Signature mirrors Hono's `streamSSE` exactly so it is a drop-in.
- All 6 existing SSE endpoints now import `streamSSE` from the wrapper: `GET /api/dump/keys/:keyId/stream` (control-plane), Anthropic Messages, Gemini `streamGenerateContent`, OpenAI Responses, OpenAI Chat Completions, and the passthrough SSE path (`/v1/completions` and friends).
- **ESLint enforcement** — a `no-restricted-imports` rule bans direct imports of `streamSSE` from `hono/streaming` inside `packages/gateway/**`; the wrapper file itself is in `ignores`. The ban uses `paths[importNames:['streamSSE']]` so sibling helpers like `streamText` and the raw `stream` helper are unaffected, and the type `SSEStreamingApi` can still be imported directly from `hono/streaming`. The common `no-restricted-imports` patterns (deep-import + platform-target bans) are extracted into a top-level `commonRestrictedImportPatterns` const so re-listing them in each override block stays a spread, not a hand-sync copy.
- **Tests** — new `streaming/sse_test.ts` locks all three response headers (`X-Accel-Buffering`, `Content-Type`, `Cache-Control`); one end-to-end assertion added to `dump_test.ts` covers the control-plane streaming path.

## Test plan

- [x] `pnpm run lint` — clean
- [x] `pnpm run typecheck` — all packages
- [x] `pnpm --filter @floway-dev/gateway test --run` — 152 files / 1937 tests
- [x] Adversarial: swapping any respond.ts back to `import { streamSSE } from 'hono/streaming'` triggers the ESLint ban with the expected message
- [ ] Manual smoke against `pnpm run dev:node` behind nginx: response headers include all three, stream chunks flush without buffering delay

### DIFF
--- a/packages/gateway/src/streaming/sse.ts
+++ b/packages/gateway/src/streaming/sse.ts
@@ -1,8 +1,5 @@
 import type { Context } from 'hono';
-// eslint-disable-next-line no-restricted-imports -- the whole point of this wrapper is to funnel every streamSSE call through here.
 import { streamSSE as honoStreamSSE, type SSEStreamingApi } from 'hono/streaming';
-
-export type { SSEStreamingApi };
 
 // nginx (and OpenResty / Tengine / other nginx-based reverse proxies) enable
 // `proxy_buffering` by default, which holds SSE chunks until the buffer fills
--- a/packages/gateway/src/streaming/sse.ts
+++ b/packages/gateway/src/streaming/sse.ts
@@ -2,6 +2,8 @@ import type { Context } from 'hono';
 // eslint-disable-next-line no-restricted-imports -- the whole point of this wrapper is to funnel every streamSSE call through here.
 import { streamSSE as honoStreamSSE, type SSEStreamingApi } from 'hono/streaming';
 
+export type { SSEStreamingApi };
+
 // nginx (and OpenResty / Tengine / other nginx-based reverse proxies) enable
 // `proxy_buffering` by default, which holds SSE chunks until the buffer fills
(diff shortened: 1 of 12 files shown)


packages/gateway/src/streaming/sse.ts
```diff
@@ -0,0 +1,18 @@
+import type { Context } from 'hono';
```
Menci, line 1: "streaming" is a concept as the same level of "control-plane" and "data-plane"

Menci:
reopen when ready to review again

yyyr-p:
Superseded by #234. The wrapper now lives at `packages/gateway/src/shared/sse.ts` (alongside `is-production-request.ts`), addressing the review feedback that "streaming" is not an architectural layer at the same level as `control-plane` / `data-plane`.

PR #229, opened 2026-07-16
feat(gateway): admit non-Anthropic models to Claude Code CLI picker

yyyr-p:
## Summary

Claude Code CLI's `/model` picker in `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1` mode [ignores every discovered id that doesn't begin with `claude` or `anthropic`](https://code.claude.com/docs/en/llm-gateway-protocol#model-discovery), so gateway-discovered non-Anthropic models (GPT-5, Gemini, custom, …) never reach the menu even though the gateway serves them fine on `/v1/messages`.

- `toClaudeCodeShape` prepends a `claude-code:` synthetic prefix on any id that doesn't already pass the CLI's `/^(claude|anthropic)/i` filter; `display_name` is preserved verbatim, and the picker renders `display_name ?? id`, so the operator-configured label is what the user sees.
- `enumerateModelCandidates` learns a symmetric strip-retry so `claude-code:<real-id>` coming back in on `/v1/messages` (or any other data-plane endpoint) resolves to `<real-id>`. Unconditional (no UA gate): the CLI's discovery UA `claude-code/*` and its Messages UA `claude-cli/*` diverge, and non-CC clients that reuse the discovered list benefit from the same symmetry.
- Same handler additionally drops embedding / image models from the CC-shape response — the picker is chat-only, matching the `.kind === 'chat'` narrow that `data-plane/codex/models.ts` and `data-plane/models/gemini.ts` already apply for their own discovery surfaces.

The predicate is extracted from `@anthropic-ai/claude-code@2.1.211`'s compiled `Bootstrap Gateway /v1/models` handler and cross-checked against the official protocol docs; both sources cited in the code.

## Test plan

- [x] `pnpm --filter @floway-dev/gateway run test` — new `serve_test.ts` case pins the prefix rewrite + `display_name` preservation + `[1m]` composition + chat-only narrow (embedding + image ids absent from CC response); new `registry_test.ts` cases pin the strip-retry (real-id resolves, unknown prefixed id misses cleanly).
- [x] `pnpm run lint`
- [x] Manual: `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1 ANTHROPIC_BASE_URL=<floway>/ ANTHROPIC_AUTH_TOKEN=<key> claude` — `/model` shows non-Anthropic chat models under their real `display_name`; embedding/image models are not listed; picking a non-Anthropic model and sending a prompt returns a response (proves inbound strip end-to-end).


packages/gateway/src/data-plane/models/claude-code-prefix.ts
```diff
@@ -0,0 +1,53 @@
...
+//    the synthesized id can never coincide with a built-in family
+//    string.
+//
+// The prefix is a Floway-owned convention: `claude-code:` was chosen
+// over `claude-` because a bare `claude-` prefix would masquerade a
+// non-Anthropic model as a Claude family member in any log / trace /
+// error the user sees; the `code:` segment plus the colon separator
+// make it visually obvious to anyone reading the wire that the id is
+// a gateway alias, not an upstream-native Anthropic name.
+export const CLAUDE_CODE_SYNTHETIC_PREFIX = 'claude-code:';
```
Menci: on hold since this PR manipulated the model resolution

yyyr-p:
Superseded by #483.

PR #231, opened 2026-07-17
feat(web): open the Claude Code / Codex settings generators to every chat model

yyyr-p:
## Summary

Opens the Claude Code and Codex settings generators to every chat model, so operators can route non-family ids through Floway's translation layer while the CLI's native family (`claude-*`, `gpt-5*`) still lands at the top of each tier slot by default. Each `<select>` splits into `<optgroup label="Claude|Codex">` and `<optgroup label="Other">`; empty groups collapse via `v-if`.

Extracts the picker's pure logic (tier-distance sort, family partition, context-window map, `[1m]` suffix) into a sibling `cli-snippet-helpers.ts` with a co-located unit test.

Sorting stays hijack-safe: `claudeTier` gates on `CLAUDE_RE` before checking tier tokens, so a non-Claude id whose name embeds a tier word (e.g. `vendor/gpt-4-opus-finetune`) cannot win the Opus slot via the reversed-`localeCompare` tiebreak.

## Test plan

- [x] `pnpm --filter @floway-dev/web run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm --filter @floway-dev/web exec vitest run src/components/keys/` (19 tests)
- [x] Load `/dashboard/keys` and confirm each select lists every chat model with the native family above `Other` under labeled `<optgroup>`s; foreign ids remain selectable.
- [x] Copy the generated `settings.json` `env` block into `~/.claude/settings.json` and confirm Claude Code CLI accepts it, including the `[1m]` suffix on any 1M-context Sonnet.
- [x] Copy the generated Codex TOML + Unix credential script and confirm `codex` picks up the Floway provider.


Menci:
Closed in favor of #168

PR #240, opened 2026-07-20
[Codex] Add ultra level support.

moooyo:
## Summary
Ultra is a client side behaviour but need backend support. The model must report it has ultra effort level and send to client side. Client will redirect it to max level and use subagent more proactively.

https://github.com/openai/codex/blob/2deed3fb9c00c74dac3d177ea700d6fb7a94539d/codex-rs/core/src/client.rs#L175-L180

so, add a fake ultra level for all gpt models which support max level. And add a settings to control this behaviour.

## Demo

Settings:
image

Client:
image


Menci:
It doesn't make sense. I see no point to have a PR with > 5 lines of change (excluding tests).

Menci:
Feel free to re-open when ready to review again.

PR #241, opened 2026-07-20
feat(codex): enable Ultra reasoning for compatible clients

### LINKED ISSUE #238: [Codex] Ultra level support.
I'm not sure, but seems ultra is just a switch to control whether use subagent or not.

Can we add a fake ultra level in response and redirect it to max(or other effor level)? to make codex desktop can proactively create subagents to improve running speed/reduce cost.

moooyo:
## Summary
Based on our discussion in https://github.com/Menci/Floway/issues/238 , we can add a fake ultra level for codex desktop (maybe also works for codex cli) to make client get more proactively subagent use and can defer agent to other models (which may reduce the cost and improve the speed).

Closed: #238 

## Detail
1. Add ultra levels for all gpt models which support max effort.
2. Fix the codex UA match logic. The real codex desktop app will send a "Codex Desktop" like UA.

## Test

Tested locally. It works. Please pay attention, this PR removed the UI changes (support ultra settings). If you think it's neccessary, please ping me I will create another PR to support it.
image

### DIFF
--- a/packages/gateway/src/data-plane/codex/catalog.ts
+++ b/packages/gateway/src/data-plane/codex/catalog.ts
@@ -22,11 +22,27 @@ import bundledCatalog from './catalog/bundled.json' with { type: 'json' };
 
 export interface CatalogModel {
   slug: string;
+  multi_agent_version?: string | null;
+  supported_reasoning_levels?: CodexReasoningLevel[];
   [key: string]: unknown;
+}
+
+export interface CodexReasoningLevel {
+  effort: string;
+  description: string;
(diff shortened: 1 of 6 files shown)


moooyo:
@Menci does this PR make sesnse? Let me know any concern or suggestions.

Menci:
I just found it's not that easy. For compatibility with older clients, a check for whether the official model catalog has ultra/v2 is introduced, to avoid exposing ultra as a real reasoning effort to the clients. Anyway, thanks!

PR #246, opened 2026-07-20
fix(responses): align advanced item wire contracts

Menci:
## Summary

Align the Responses protocol types with the live upstream wire contracts for
the advanced output/input items, and model the queued lifecycle as a
first-class response envelope.

- Replace the permissive placeholder shapes for computer / computer_use_preview,
  tool-search call/output, code-interpreter, local-shell call/output,
  apply-patch, file-search, and MCP call/list-tools/approval items with their
  current live Responses contracts. Each vendor constant carries a pinned
  upstream reference, and the input/output optionality that round-tripped
  continuation items require is preserved.
- Keep strict payload-free request controls (e.g. `compaction_trigger`) distinct
  from output items, so item-id readers no longer invent an id slot; the
  client-output pipeline is typed as Responses output items.
- Drop the standalone `compaction_summary` type, retaining only the Codex
  deserialize alias onto `compaction`.
- Complete the built-in `tool_choice` union so every declared built-in tool can
  be forced, guarded by compile-time fixtures.
- Treat `response.queued` as an official response envelope across stream
  parsing, affinity wrapping, client response-id rewriting, retry prologue
  detection, and the server-tool shim, so its nested output cannot bypass the
  identity membranes. Reject the nonterminal `queued` / `in_progress` /
  `cancelled` statuses in the terminal builders, so a queued or cancelled
  snapshot can never be mis-emitted as `completed`.

## Test Plan

- [x] `pnpm run test` — 380 files pass
- [x] `pnpm run lint`
- [x] `pnpm run typecheck`

## Stack

1 of 4. The next PR moves gateway-created items to producer-owned random IDs.

### DIFF
--- a/packages/gateway/src/data-plane/chat/responses/affinity/egress.ts
+++ b/packages/gateway/src/data-plane/chat/responses/affinity/egress.ts
@@ -82,7 +82,8 @@ const wrapNaturalResponsesAffinity = async function* (
       continue;
     }
     if (
-      event.type === 'response.created'
+      event.type === 'response.queued'
+      || event.type === 'response.created'
       || event.type === 'response.in_progress'
       || event.type === 'response.completed'
       || event.type === 'response.incomplete'
@@ -268,7 +269,7 @@ const wrapResponsesFirstCarrier = async function* (
       return;
     }
 
-    if (event.type === 'response.created' || event.type === 'response.in_progress') {
+    if (event.type === 'response.queued' || event.type === 'response.created' || event.type === 'response.in_progress') {
       yield* discoverFirstFromSnapshot(event.response, event.sequence_number);
       const response = await rewriteResponse(event.response, false);
       yield eventFrame(addSequenceOffset({ ...event, response }, sequenceOffset));
--- a/packages/gateway/src/data-plane/chat/responses/affinity/egress.ts
+++ b/packages/gateway/src/data-plane/chat/responses/affinity/egress.ts
@@ -269,7 +269,13 @@ const wrapResponsesFirstCarrier = async function* (
       return;
     }
 
-    if (event.type === 'response.queued' || event.type === 'response.created' || event.type === 'response.in_progress') {
+    if (event.type === 'response.queued') {
+      const response = await rewriteResponse(event.response, false);
+      yield eventFrame(addSequenceOffset({ ...event, response }, sequenceOffset));
+      continue;
+    }
+
+    if (event.type === 'response.created' || event.type === 'response.in_progress') {
       yield* discoverFirstFromSnapshot(event.response, event.sequence_number);
       const response = await rewriteResponse(event.response, false);
       yield eventFrame(addSequenceOffset({ ...event, response }, sequenceOffset));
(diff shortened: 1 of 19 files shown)


Menci:
Codex agent — live wire evidence for the computer-call split: Azure GPT-5.4 emitted the modern `computer` call with `actions` and rejected requests that also carried `pending_safety_checks`, including an empty array. The legacy `computer_use_preview` generation used singular `action` with `pending_safety_checks`. The generated SDK/OpenAPI schema currently exposes both fields in one permissive shape, so the mutually exclusive Floway union follows observed service validation rather than mechanically copying that generated shape.

Menci:
Codex agent — final research record for this PR.

# Research: Responses wire contracts

## Scope and decision

This PR repairs the shared Responses type model before changing item identity.
It does not introduce an ID policy. Request controls remain input-only, output
items model the fields actually emitted by current Responses implementations,
and `response.queued` cannot suppress fast-path expansion or bypass the audited
response/blob membranes.

`context_compaction` and `compaction_summary` remain explicit legacy wire
compatibility variants. Current compaction is `compaction_trigger` input to
`compaction` output.

## Findings

- `file_search_call` requires `queries`, `status`, and optional structured
  `results`; the old permissive shape hid field drift.
- Modern computer calls use `actions`; legacy computer preview calls use one
  `action` plus safety checks. Their mutually exclusive wire generations and
  all nine action variants are modeled explicitly. Computer output is a
  screenshot object.
- `code_interpreter_call` carries `code`, `container_id`, structured outputs,
  and status; it has no function-style `call_id`.
- Local shell and modern shell are distinct protocols. Local shell output is
  opaque text; modern shell output is structured stdout/stderr/outcome chunks.
- `tool_search_call` uses `arguments`; `tool_search_output` uses `tools`.
- Program, program-output, agent-message, multi-agent, and compaction fields
  were checked against current SDK/Codex definitions.
- Apply-patch uses structured create/delete/update operations. MCP calls and
  approval items do not share function-call field names.
- `response.queued` is an official response-envelope event. Treating unknown
  events as content meant a queued wrapper could suppress fast-path expansion
  and bypass blob/response-ID processing, so it is modeled across parser,
  affinity, retry, hosted-tool, and client-output boundaries without emitting
  output before the queued envelope.
- Resource status also includes `cancelled`. JSON-to-stream and translated
  terminal builders reject queued, in-progress, and cancelled resources rather
  than silently converting them into successful completion.
- Tool-search results use the complete current tool declaration union, and the
  matching built-in forced-tool choices are represented alongside them.

## Primary references

- OpenAI Node Responses definitions:
  https://github.com/openai/openai-node/blob/39a15b412fc129df15339ebd6e3e6547854aa81f/src/resources/responses/responses.ts
- OpenAI OpenAPI examples and prefixes:
  https://github.com/openai/openai-openapi/blob/db3e53198a66732cfe161339ea63bf36fc0137ad/openapi.yaml
- Codex Responses variants and prefixes:
  https://github.com/openai/codex/blob/8c41ed33ce3e39460e7b13b14c35e0c39bb5980d/codex-rs/protocol/src/models.rs#L1060-L1100
- Local shell behavior:
  https://github.com/openai/openai-agents-python/blob/2fa463571e76dae8ff267622f1018eaf06ffeb9f/tests/test_local_shell_tool.py#L46-L92

## Verification

- `pnpm run test -- --reporter=dot`: 380 files, 4620 tests
- `pnpm run lint`
- `pnpm run typecheck`

PR #247, opened 2026-07-20
refactor(responses): mint item IDs at producer boundaries

Menci:
## Summary

Every Responses item the gateway itself synthesizes now mints a final, random
producer ID at the point it is created, replacing the positional placeholders
(`rs_0`, `msg_0`, `fc_0`) and the ad-hoc `_gw`/UUID schemes.

- The Messages→Responses and Chat-Completions→Responses translators, plus the
  shared reasoning projections, mint an item's ID when it opens.
- Hosted tools (web search, image generation), the server-tool shim, the
  affinity reasoning prefix, compact output, and retained compact messages all
  receive producer-owned random IDs at creation.
- One ID stays stable across a stream item's `added`, child/delta, `done`, and
  terminal-snapshot frames.
- A single `createRandomResponsesItemId` helper owns the canonical prefixes and
  hard-fails on unsupported runtime item types; reasoning fallback IDs are drawn
  only when the source item genuinely lacks one.
- With every synthetic producer on a normal producer ID, the orphaned `_tmp_`
  marker and the `outputItemSource` guard that recognized it are removed. The
  affinity reasoning prefix stays excluded from upstream attribution because
  ingress always strips the originless carrier before candidate-time restoration
  could read it.

The state-writing client membrane may still alias these producer IDs; #249
removes that membrane, at which point they become the client-visible IDs.

## Test Plan

- [x] `pnpm run test` — 382 files pass
- [x] `pnpm run lint`
- [x] `pnpm run typecheck`

## Stack

1 of 3 remaining. #246 was squash-merged into `main`; #248 is stacked next.

### DIFF
--- a/TRANSLATION.md
+++ b/TRANSLATION.md
@@ -416,7 +416,7 @@ Request mapping:
   unpacked from the `${encrypted_content}@${id}` shape this gateway emits: the
   Responses reasoning id and any opaque `encrypted_content` are recovered. A
   native signature carrying no `@` is preserved verbatim as `encrypted_content`
-  with a synthesized `rs_${index}` id; it is never overwritten.
+  with a fresh random `rs_` id; it is never overwritten.
 - `max_tokens`, `temperature`, `top_p`, `metadata`, and `stream` pass through
   when present.
 - `output_config.effort` maps directly to `reasoning.effort`; disabled thinking
@@ -497,7 +497,7 @@ Response mapping:
 - Responses output items are converted in output order.
 - `reasoning` maps to a Messages thinking carrier; the upstream's genuine
   `signature` (or `redacted_thinking` `data`) is carried verbatim as the
-  reasoning item's `encrypted_content`, with a synthesized `rs_${index}` id.
+  reasoning item's `encrypted_content`, with a fresh random `rs_` id.
 - `message` content maps to text. `refusal` content is kept visible as text
   because Messages has no local refusal block.
 - `function_call` maps to `tool_use`.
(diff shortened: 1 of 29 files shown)


Menci:
Codex agent — final research record for this PR.

# Research: producer-owned synthetic Responses IDs

## Scope and decision

Every component that creates a Responses item now assigns its stable producer
ID at construction time. Output indexes remain ordering keys, not identity.
The current state-writing client membrane may still alias producer IDs before
they reach the client; that membrane is removed by the final stacked PR.
Native Responses providers are untouched here.

Generated IDs use the item-specific prefix plus 16 random bytes rendered as 32
lowercase hex characters. Hex avoids Copilot's observed rejection of a trailing
base64url underscore while keeping IDs fixed-length and opaque. The exported
generator rejects unknown runtime types before drawing randomness; reasoning
fallback IDs are drawn lazily only when the source lacks an ID.

## Producers covered

- Responses translated from Chat Completions: message, reasoning, function,
  and custom-tool output lifecycles.
- Responses translated from Messages: message, reasoning, function, and
  custom-tool output lifecycles.
- Hosted web search and image generation.
- Synthetic affinity reasoning prefixes.
- Compact-shim compaction items.
- Server-tool merged messages when an inner stream has no ID.
- Retained messages synthesized into Copilot's compact value envelope.

One ID is allocated when the item opens and reused by added, child, done, and
terminal frames. A carried upstream reasoning ID is still preserved when it is
real state rather than a newly synthesized item.

## Prefix evidence

- OpenAI OpenAPI examples use `msg`, `rs`, `ws`, `fc`, `ctc`, and `cmp`:
  https://github.com/openai/openai-openapi/blob/db3e53198a66732cfe161339ea63bf36fc0137ad/openapi.yaml#L57042-L59599
  https://github.com/openai/openai-openapi/blob/db3e53198a66732cfe161339ea63bf36fc0137ad/openapi.yaml#L68023-L68281
- Current Codex confirms the common mapping, including `ig`:
  https://github.com/openai/codex/blob/8c41ed33ce3e39460e7b13b14c35e0c39bb5980d/codex-rs/protocol/src/models.rs#L1076-L1094

The shared generator intentionally contains only types Floway creates. It is
not a catalog or validator for native upstream IDs; Copilot's broader output
inventory remains provider-private in the next PR.

## Verification

- `pnpm run test -- --reporter=dot`: 382 files, 4639 tests
- `pnpm run lint`
- `pnpm run typecheck`

PR #248, opened 2026-07-20
refactor(copilot): isolate Responses item identity

Menci:
## Superseded by #252

The original squash commit for this PR was removed from `main` by a later force-push. The identical diff, PR description, and research comment were rebuilt and squash-merged in #252 onto the restored linear history.

Continue at #252.


Menci:
Codex agent — final research record for this PR. Probe summaries intentionally omit credentials, raw opaque blobs, and account-specific endpoints.

# Research: Copilot Responses item identity membrane

## Decision and upstream constraints

Copilot Responses behaves as a `store:false` upstream: replayable state must be carried by the client. A raw Copilot item ID is neither a durable handle nor a routing signal by itself. The provider therefore exposes a fresh type-correct public ID for every verified output item, and restores a raw upstream ID only when that ID is paired with provider-owned replay state carried inside the same item.

The provider boundary is independent from generic affinity. Affinity may append its own authenticated outer metadata and may strip that layer when routing changes; Copilot then treats the remaining item as foreign unless its own inner trailer is still present.

## Carrier inventory and semantics

Carrier behavior is type-specific:

| output item | public prefix | carrier | evidence and replay behavior |
|---|---:|---|---|
| `reasoning` | `rs_*` | `encrypted_content` | Under Azure GPT-5.4 source `store:false`, replay succeeded only with the exact `(raw rs_* ID, encrypted_content)` pair. Changing only the ID deterministically produced `item_id mismatch`. |
| `program` | `cm_*` | `fingerprint` | Copilot GPT-5.6 PTC produced it. `fingerprint` is the only carrier; adding top-level `encrypted_content` is invalid. |
| `compaction` | `cmp_*` | `encrypted_content` | Current Copilot canonical item. The blob itself accepted a fresh legal `cmp_*` ID and still hit cache, but the membrane uniformly preserves the exact same-frame pair. |
| `agent_message` | `amsg_*` | nested `content[].encrypted_content` | This is Codex multi-agent wire, not ordinary Copilot assistant output. A top-level encrypted field is invalid. Current Copilot multi-agent beta probes did not emit it. |

The missing-ID question was explicitly rechecked. Fresh Copilot probes observed two carrier-bearing reasoning `.added` frames and one program `.added`; all had nonempty IDs. Historical dumps contained 11 carrier-bearing compaction `.added` observations; all 11 had nonempty IDs and blobs. No carrier-bearing output without an ID was found. Passing such a blob with a random public ID would recreate the signature mismatch, while silently stripping it would discard state, so the membrane intentionally fails if a future upstream violates this contract.

## Carrier-free output inventory

The carrier-free whitelist came from complete-item echo trials, not prefix inference:

| output item | public prefix | fresh-ID replay | constraint |
|---|---:|---|---|
| `message` | `msg_*` | 200, cache-neutral | complete literal |
| `function_call` | `fc_*` | 200, cache-neutral | preserve `call_id` |
| `custom_tool_call` | `ctc_*` | 200, cache-neutral | paired output remains separate |
| `web_search_call` | `ws_*` | 200, cache-neutral but lossy | result list is not self-contained |
| `tool_search_call` | `tsc_*` | 200, cache-neutral | prefix verified live |
| `tool_search_output` | `tso_*` | 200; raw/fresh equally lossy | server execution emits this item without a carrier |
| `program_output` | `cmo_*` | 200, cache-neutral | Copilot GPT-5.6 PTC |
| `shell_call` | `sh_*` | 200, cache-neutral | added/done raw IDs differed |
| `shell_call_output` | `sho_*` | 200, cache-neutral | random and omitted IDs both accepted |
| `apply_patch_call` | `apc_*` | 200, cache-neutral | prefix verified live |

Two Copilot requests rejected base64url-style `ws_*` IDs ending in `_`; hex suffixes succeeded. Public IDs therefore use 16 random bytes rendered as 32 lowercase hex characters.

A final server-side tool-search probe exposed a previously missing live output kind. With `stream:true, store:false`, gpt-5.4 emitted `tool_search_call` and `tool_search_output` added/done frames. The output keys were exactly `call_id, execution, id, status, tools, type`; it had no blob carrier, and its raw ID was a 408-character unprefixed opaque value. Follow-ups with raw IDs, fresh legal `tsc_*/tso_*` IDs, and omitted IDs all returned HTTP 200/completed. Two raw and two fresh-ID repeats were equally unable to recall a known tool name, while one omitted-ID trial exposed the literal tool list. Fresh public IDs were therefore no worse than raw opaque IDs under `store:false`, and the type belongs in the carrier-free policy.

## Cache experiments

Cache trials changed only legal item IDs while leaving a long stable suffix unchanged. The changed IDs appeared before tens of thousands of stable tokens; one apply-patch trial changed 80 IDs. Warm replays retained the same cached-token depth, reaching approximately 94k cached tokens. Changing actual message/tool/action/output content, unlike changing IDs, moved or destroyed the cache boundary. This is why carrier-free ID randomization is considered cache-neutral rather than merely request-valid.

## Azure `store:false` reference matrix

The Azure reference used GPT-5.4 items produced by a source request with `store:false`, then compared a complete original-ID echo with an otherwise identical fresh legal ID.

Equivalent after ID replacement: message, function call, custom tool call, tool search call, computer call, code interpreter call, shell call/output, apply patch call, MCP list tools, MCP approval request when `approval_request_id` changed consistently, and compaction.

Equally lossy or failing: web search call (result list unavailable), MCP call (accepted but absent from model context), and image generation call (404). Only reasoning was non-equivalent because its blob was cryptographically ID-bound. Every comparison that reached inference was cache-neutral. Azure `store:true` known-handle behavior was deliberately not used as a Copilot constraint.

The closed Copilot output policy remains narrower than the Azure protocol inventory. Counterexamples justify failing closed: a fresh image-generation ID remained a 404 even with a full image result; a fresh MCP call was accepted but disappeared from context; randomizing an MCP approval request can break approval pairing; and a new item may place state in a non-obvious slot such as `program.fingerprint`. Prefixes also cannot be inferred from type names (`computer → cu`, `tool_search → tsc`). Unknown Copilot output kinds therefore fail before yielding a raw ID; foreign client input is not subject to that output whitelist.

## Web-search recall experiment

Two blinded experiments selected target fields only after the first turn, excluded facts already present in the visible answer/reasoning summary, and disabled tools on follow-up.

Run 1 used five independent searches and four branches per search: raw `ws_*` plus reasoning, randomized `ws_*` plus reasoning, reasoning-only, and web-item-only. Across 20 continuations and 100 previously unmentioned requested fields, 100/100 were not recovered, including result count.

Run 2 used result lists of 35/40/25/37 entries and targeted entries 26/37/23/34 across 16 continuations. Search-list recovery was 0/16, reasoning-blob branches were 0/8, and raw-ID branches were 0/8. A visible-assistant-history control succeeded 16/16, proving that follow-up history itself arrived. Neither raw ID, literal web-search item, nor post-search reasoning reconstructed unseen search results.

## Streaming lifecycle evidence

A live Copilot compaction stream placed `encrypted_content` on `output_item.added`, and its added, done, and terminal observations used different raw IDs. Therefore every carrier observation appends its own same-frame raw ID to its own opaque value. Replacing an earlier frame with later done/terminal identity would invalidate the earlier blob.

`responsesResultToEvents` strips reasoning blobs from synthesized added events, but program, compaction, and agent-message generic paths may carry replay state at added. Normal Codex persistence stores done items, but a `store:false` client may replay an added item directly, so added-frame correctness remains part of the wire contract. The membrane allocates a public ID at first observation without buffering and reuses it across child, added, done, and response snapshot frames.

Snapshot-first tracking was also checked. Copilot's observed created/in-progress snapshots were empty, but a nonempty snapshot would create `{publicId, added:false}`; a later added frame reuses the public ID and flips only the lifecycle flag. It is not mistaken for a duplicate added event.

## Final direct stream probes

All final probes called Copilot `/responses` directly with `stream:true, store:false`.

- Ordinary reasoning used `reasoning:{effort:'high',summary:'detailed'}` and `include:['reasoning.encrypted_content']` on gpt-5.3-codex, gpt-5.4, gpt-5.5, and gpt-5.6-sol. None emitted `response.reasoning_text.{delta,done}`; gpt-5.5 also omitted reasoning-summary events. The reasoning-text events are nevertheless official OpenAI leaf variants carrying `item_id + output_index`, so the membrane models them defensively.
- Shell used `tools:[{type:'shell',environment:{type:'local'}}]` with forced shell choice. `response.shell_call_command.{added,delta,done}` had `output_index` and no `item_id`; adding `environment.os_type` was rejected.
- Apply patch used `tools:[{type:'apply_patch'}]` with forced apply-patch choice. Both operation-diff events carried `item_id` and `output_index`.
- Tool search required at least one `defer_loading:true` tool. Server execution emitted call and output items; client execution emitted the call item. Neither emitted tool-search child deltas.
- Fresh GPT-5.6 PTC produced program added with both ID and fingerprint. Fresh multi-agent beta requests did not cause Copilot to emit agent messages.

The configured production proxy used a curl-unsupported scheme, so these direct probes followed the probing procedure's direct-egress fallback rather than the production proxy path. The enterprise catalog advertised `/responses` for gpt-5-mini, gpt-5.3-codex, gpt-5.4, gpt-5.4-mini, gpt-5.5, gpt-5.6-luna/sol/terra, and mai-code-1-flash-picker; event claims above apply only to the successfully tested models.

## Trailer and boundary order

The provider appends:

```text
original decoded/raw blob bytes
+ UTF-8 JSON {"version":1,"origin":"raw|base64|base64url","id":"..."}
+ uint16be(JSON byte length)
```

No account identity or encryption is included. The trailer is not authorization or routing state, exposes only the already client-visible opaque upstream ID, and is independently verified by Copilot when replayed. Generic affinity appends its own authenticated outer layer after this provider layer. On replay, affinity removes/projects its outer layer first; Copilot then removes its inner layer. Double append is intentional and the two metadata formats remain independent.

Future leaf stream events are handled structurally: a top-level `item_id` requires numeric `output_index` and is rewritten; an identity-free leaf passes through. Unknown envelopes containing a complete `item` or `response` fail closed. This covers all 53 official OpenAI event variants: currently unmodeled variants are either identity-free audio leaves or `item_id + output_index` leaves. The closed output-item whitelist remains a separate safety boundary.

## References

- Official Responses item examples and prefixes: https://github.com/openai/openai-openapi/blob/db3e53198a66732cfe161339ea63bf36fc0137ad/openapi.yaml#L68023-L68748
- Agent-message example: https://github.com/openai/openai-openapi/blob/db3e53198a66732cfe161339ea63bf36fc0137ad/openapi.yaml#L74970-L75020
- Codex prefix inventory: https://github.com/openai/codex/blob/8c41ed33ce3e39460e7b13b14c35e0c39bb5980d/codex-rs/protocol/src/models.rs#L1076-L1094
- Official Responses stream-event union: https://github.com/openai/openai-node/blob/39a15b412fc129df15339ebd6e3e6547854aa81f/src/resources/responses/responses.ts#L6875-L6929
- Official reasoning-text delta/done shapes: https://github.com/openai/openai-python/blob/d4dceb221b9a92c55c232d5b330ae89beb539415/src/openai/types/responses/response_reasoning_text_delta_event.py#L9-L31 and https://github.com/openai/openai-python/blob/d4dceb221b9a92c55c232d5b330ae89beb539415/src/openai/types/responses/response_reasoning_text_done_event.py#L9-L34

## Verification

- Real provider plus generic-affinity inner/outer round-trip integration test
- Direct Copilot stream and replay probes across current GPT model families
- `pnpm run test -- --reporter=dot`: 384 files, 4678 tests
- `pnpm run lint`
- `pnpm run typecheck`

PR #249, opened 2026-07-20
refactor(responses): rebuild stateful persistence

Menci:
## Summary

Rebuild Stateful Responses persistence around producer-owned item identity and opt-in, whole-day retention.

- Preserve native upstream item IDs at the client boundary. Copilot and gateway-owned synthetic producers allocate protocol-valid IDs; the generic gateway no longer rewrites item identity. Response envelope IDs remain opaque and are generated once at the source boundary for the complete client response.
- Persist the first complete client-facing item synchronously at `response.output_item.done`, then commit the response snapshot at the successful terminal event without replacing later wire content.
- Add `responses_retention_seconds` per API key, defaulting to `0`. Positive values remain seconds on the wire and in storage, shared with the dump duration representation, but must be whole-day multiples from 1 through 3,650 days.
- Expose Off, 7-day, 30-day, and custom integer-day controls in the API-key editor. The copy distinguishes HTTP durable persistence from Codex's independent non-persistent WebSocket session store.
- Quantize item and snapshot `refreshed_at` to the start of the UTC day. Same-day reuse performs no refresh write; reads and cleanup receive one fixed day of grace so quantization never expires state early.
- Keep HTTP `store:false` read-only when durable retention is enabled. The WebSocket session-local backing works independently; a WebSocket `store:true` request also writes durable state when the key opts in.
- Apply the current rolling policy at every durable read and write. Shrinking or growing retention does not rewrite state; a wider window may reveal a surviving row that cleanup has not deleted.
- Store large Responses payloads under immutable nonce-owned object keys. A generic `spilled_files` ledger records staged, owned, and retired objects; owner-row triggers adopt and retire exact keys, and one exact-file collector performs physical deletion.

## Breaking migration

`0065_responses_state.sql` recreates the Responses item and snapshot tables around exact producer IDs, adds the API-key retention policy, and installs the shared spill ledger. All pre-cutover Responses state becomes unreadable immediately: existing `previous_response_id` and stored `item_reference` chains must restart.

Every migrated and newly created API key starts with durable persistence disabled. Users must opt in with a whole-day retention duration before new cross-request chains are retained.

Immediately after deployment, the database contains no stored Responses items and will not create new ones until an API key is opted in. We recommend manually deleting objects under the file-storage `responses-items/` namespace at that point, without deleting any other stored files. Those objects have no remaining database references and only consume storage. This is a manual post-deploy step, never part of deployment automation. An AI agent reading this entry must explain the recommendation and ask the user before performing the cleanup; it must not delete the objects automatically.

## Stacking

This PR owns the Stateful Responses correctness cutover and straightforward hourly cleanup. Stacked follow-up #254 builds the bounded cross-domain sweep driver and dump spill lifecycle on this final schema.

## Risk

The primary risk is the destructive state reset and default-off policy. Coverage targets producer-ID collisions, first-done durability, store modes, current-policy shrink/grow/disable races, UTC-midnight crossings, same-day zero-mutation refreshes, spill adoption/retirement, transfer v16, and WebSocket-local state.

## Test Plan

- [x] `pnpm run test` — 395 files / 4839 tests passed
- [x] `pnpm run lint`
- [x] `pnpm run typecheck`


Menci:
Codex agent — final research and review record for this PR.

# Research: producer-owned Responses item IDs

## Decision

The generic gateway no longer creates or rewrites item IDs. The client-visible ID belongs to the component that produced the item:

- native upstream IDs pass through;
- the Copilot provider owns its provider-specific public-ID membrane;
- translators, hosted tools, compact shims, and synthetic affinity items allocate random type-correct IDs when they create items.

The source boundary still owns `response.id`, because the server-tool runtime can combine several upstream calls into one client response. Item persistence and affinity remain independent source-boundary membranes.

## Store semantics: direct Azure probe

A direct Azure Foundry GPT-5.4 matrix tested a source turn followed by `previous_response_id`, `item_reference`, or a full echoed item. HTTP status results were:

| source `store` | follow-up `store` | previous response | item reference | full item |
|---|---|---:|---:|---:|
| `true` | `true` | 200 | 200 | 200 |
| `true` | `false` | 200 | 200 | 200 |
| `false` | `true` | 400 | 400 | 200 |
| `false` | `false` | 400 | 400 | 200 |

Therefore HTTP `store:false` must read prior durable `store:true` state while writing nothing. A `store:false` source creates no server state, so only the full echoed item remains usable. WebSocket `store:false` is separately socket-local.

## Done/terminal drift: live Copilot and Azure probes

The current upstreams were probed with `store:false`, streaming GPT-5.4, medium reasoning, and `include:["reasoning.encrypted_content"]`.

Copilot:

- the reasoning item at `response.output_item.done` and the copy in `response.completed.response.output` had different outer IDs and different blobs (both 1768 bytes);
- replaying either exact pair returned 200;
- cross-pairing done ID + terminal blob and terminal ID + done blob also returned 200;
- corrupting one blob character returned 400;
- both short follow-ups reported `cached_tokens: 0`.

Azure Foundry:

- done and terminal kept the same `rs_…` ID but had different blobs (both 1144 bytes);
- either exact carrier replayed with 200;
- corrupting one character returned 400 with encrypted-content verification/decryption failure;
- both short follow-ups reported `cached_tokens: 0`.

Neither carrier alone recovered a calculation present only in hidden reasoning. This proves each blob is independently validated replay state, not that it is standalone conversation memory; `cached_tokens: 0` on these short probes also does not prove caching can never apply.

The Copilot probe called `api.*.githubcopilot.com` directly with the production account token. Its configured production egress is `ss://`, which curl could not mirror; the Azure probe was direct to the configured Foundry endpoint.

## Why the generic canonicalizer stays deleted

[#119](https://github.com/Menci/Floway/pull/119) introduced generic ID/blob pinning after raw Copilot streams exposed different done/completed IDs and the old gateway-ID mapper stored two rows. Its own replay matrix showed done/done, terminal/terminal, and both cross-pairs succeeding for reasoning and compaction. The blob pin was additionally needed by the then-current ciphertext-hash affinity lookup; it was not an upstream protocol requirement.

That old mapper and blob-hash affinity consumer no longer exist. Codex consumes done items while OpenAI SDK/LiteLLM paths consume terminal output, so there is no universal industry frame to substitute for the other. [Commit `f9161c8`](https://github.com/Menci/Floway/commit/f9161c84016aa3f5c79e9fc3d199dd34ad3d8996) consequently chose first-done durability while preserving every later upstream frame. The new live probes independently confirm that decision.

Copilot's lifecycle ID drift is handled inside the Copilot provider, not at the client-facing persistence boundary: its public random ID remains stable while each blob carries that frame's raw Copilot ID. Unknown Copilot output types fail before a raw ID leaks. Other providers are responsible for emitting protocol-valid IDs.

## Persistence model

- At the first `response.output_item.done`, the exact post-affinity client item and optional server-private payload commit before the frame is published.
- Repeated done and terminal items remain untouched. The terminal event commits only the ordered snapshot of first-done IDs; failed/error streams keep completed item rows but no response snapshot.
- Full items and `item_reference`s hydrate by arbitrary exact ID, without prefix/checksum recognition or candidate-specific rewriting.
- A stateful idless input receives a random internal snapshot key and deduplicates by a canonical JSON content hash. HTTP `store:false` neither hashes nor stages it.
- The repository treats exact API-key/ID/item/private-payload reuse as idempotent and rejects a different row under the same scoped ID. There is no pre-affinity “stable identity”, affinity-target identity, cross-backing preflight, or refresh-time caller-payload identity.
- Response IDs are opaque `resp_` + 128 random bits; the removed CRC verifier had no runtime consumer.

## SQL and spill findings fixed during review

- Arbitrary producer IDs never enter object paths. Spill keys use the API-key-scoped SHA-256 digest of the ID plus a per-write nonce.
- Insert collision checks validate the actual persisted payload before and after `ON CONFLICT DO NOTHING`; exact duplicates only refresh lifetime.
- Lifetime refresh accepts only `{apiKeyId,id}`. It moves the persisted payload rather than trusting caller content.
- A descriptor read can race a concurrent CAS that publishes a replacement spill and deletes the old file. Both normal hydration and refresh preparation now re-read only when `payload_json` actually changed; unchanged missing/corrupt data still throws.
- A stale refresh that discovers a newer descriptor never moves it into an older expiry bucket. `MAX(created_at, ?)` and the descriptor CAS keep row lifetime and spill expiry monotonic.
- Generated loser files are deleted by nonce ownership; the current winner's spill is retained.

Keeping every old spill instead of retrying readers was rejected: an active large item can refresh into a new hourly expiry bucket and would otherwise accumulate up to 30 days of obsolete copies.

## Removed review overreach

A line-by-line review deleted the identity-reconciliation subsystem that had modeled unreachable producer-ID reuse across targets and turns: `ResponsesInputIdentity`, `stableIdentity`, `payload.identity`, pre-affinity decode, cross-backing comparison, terminal-only persistence fallback, and refresh caller-identity validation. Output ordering now belongs to the stream membrane, and HTTP `store:false` performs no persistence work. Tests were rewritten around real provider added/done/terminal lifecycles rather than terminal-only mocks.

The only extra concurrency branches retained are the descriptor-change retries above; deterministic hooks reproduce both real interleavings, and independent review found no test-only production branch.

## Migration and risk

`0065_responses_producer_item_ids.sql` deliberately drops all prior Responses items and snapshots and recreates the exact-item schema:

```text
responses_items(id, api_key_id, payload_json, content_hash, created_at)
responses_snapshots(id, api_key_id, item_ids_json, created_at)
```

No old-state compatibility path remains. Existing `previous_response_id` and stored `item_reference` chains must restart after deployment.

## Verification

- Two fresh review/cleanup passes converged with zero actionable findings on the final HEAD.
- Copilot/Azure carrier replay and Azure store-mode behavior were probed live as recorded above.
- `CI=1 pnpm exec vitest run --reporter=dot`: 393 files, 4814 tests passed.
- `pnpm run lint`: passed.
- `pnpm run typecheck`: passed.

Menci:
Codex agent — retention and D1 design record for the unified Stateful Responses cutover.

## Policy decisions

- Durable Responses persistence is an API-key policy. The default is `0`, meaning disabled: persistent item references and `previous_response_id` are not found, and the HTTP request path does not query the durable state tables at all.
- Positive values accept 1 hour through 10 years. The dashboard provides 7-day and 30-day presets plus a validated custom duration.
- HTTP `store:false` remains read-only when the key has opted in: it can consume an earlier durable chain but never writes or refreshes state. WebSocket `store:false` remains socket-local and is unaffected by the durable setting.
- A positive retention change keeps the epoch. A monotonic visibility floor preserves exactly the set still visible across a shrink, preventing cleanup timing from making older rows reappear after a later expansion. Setting the same value is a no-op. Disabling persistence rotates the epoch and resets the floor, so in-flight writers cannot make the disabled generation visible later.
- Public API-key CRUD and import/export expose only `responses_retention_seconds`; the epoch and visibility floor are internal concurrency state. Import uses compare-and-swap over all three fields so stale merges cannot re-enable or resurrect state.

## Storage and race model

- Read visibility is the intersection of the current API-key epoch and `refreshed_at >= max(now - retention, visible_after)`.
- Items and snapshots are keyed by API key plus epoch. A snapshot's refresh time is the minimum refresh time of its members, so a response cannot outlive any referenced item.
- Producer-owned item IDs are immutable within one visible API-key epoch. Exact payload reuse is idempotent; conflicting payloads reject. Replacement is allowed only after the old row is outside both the request cutoff and the current database policy.
- Refresh updates timestamps only and uses a payload-hash compare-and-swap. It rechecks current policy inside SQL, preventing a request that started before a shrink from reviving excluded history.
- First `response.output_item.done` durability remains synchronous because the client may reference that item as soon as the frame is published. Successful terminal snapshot durability is also synchronous.

## D1 overload archaeology and resulting bounds

Prior production investigation found `D1_ERROR: D1 DB is overloaded. Requests queued for too long.` around hourly Responses cleanup. The previous shape produced roughly 234k–293k row mutations at a tick, a 1,413-item response could generate more than 200 SQL statements, D1 batches share a 30-second budget, and large migration-time deletion had already produced partially applied 7009/7500 failures. These findings ruled out per-key unbounded sweeps, migration-time mass deletion, and per-item SQL loops.

The resulting maintenance design is bounded and resumable:

- Exact item lookup/refresh uses indexed, chunked statements and keeps the 1,413-item regression below the D1 statement budget.
- Current-state cleanup claims due API keys with revision tokens and processes at most three keys per tick, or two while v1 disposal is active. Each delete is a bounded indexed range.
- Pre-cutover D1 rows and the `responses-items/v1/` object root are drained by persisted cursors and one file-provider page per tick, after an old-writer grace horizon. They are never consulted by the readable state path.
- New spill objects use immutable nonce-owned keys. D1 stages exact keys before R2/file writes; triggers adopt winners and queue replaced/deleted objects. Refreshing a stable payload does not read, copy, or delete its object.
- Payload garbage collection claims at most 1,000 exact keys and deletes them with the platform batch API.

## Dump maintenance discovered in the same budget audit

The scheduled Worker shared the same D1 budget with dump cleanup, whose old full-key scans and prefix deletion could invalidate the Responses bounds. The unified change therefore makes dump cleanup bounded as well:

- Dump bodies use nonce-owned paths and exact-file staging/adoption/retirement.
- Per-key maintenance rows carry due time, revision, claim, and a persisted visibility-floor cursor.
- Retention shrink, disable, replace, and purge advance a logical visibility floor immediately; physical row/file deletion is queued and bounded.
- A late request that started before the floor transition resets the cursor, covering SQLite rowid reuse without rescanning unbounded history.
- One tick performs a bounded backfill plus at most four cleanup units; exact file GC is capped at 1,000 keys.

## Migration shape

The PR now has one migration relative to `main`: `0065_api_key_responses_retention.sql`. It creates only the final API-key policy fields, current Responses tables/indexes/triggers, bounded Responses maintenance state, and bounded dump maintenance state. The pre-cutover tables remain isolated solely so the cursor can dispose of them without an overload-prone migration transaction. The new state tables start empty and all keys start disabled.

## Verification

- Independent retention-semantics, control/UI/data-transfer, and D1/R2 payload/cleanup reviews converged with zero actionable findings before this branch consolidation.
- After folding retention into #249 and removing the intermediate migration, the full suite passed: 395 files / 4870 tests.
- `pnpm run lint` and `pnpm run typecheck` passed on the final pushed HEAD.

Menci:
Codex agent — architecture correction and daily-retention addendum. This comment supersedes every retention/sweep claim in the two earlier research comments while preserving their producer-ID, Azure store-mode, done/terminal, and spill-race evidence.

## Original decision record

Claude session `eaa733c5-10fa-41ef-b940-7457d9976881` was read through its live parent chain across compactions and rewinds.

- Line 2432 records the proposal: items used during a day should not be renewed again that day; configured retention is expressed in days while the stored value remains seconds.
- Line 2462 rejects a retention-dependent `f(retention)` throttle: the intended tradeoff is a fixed one-day refresh throttle and cleanup delayed by one day, exchanging at most one extra day of storage for substantially fewer writes.
- Lines 2465–2467 settle the design: fixed `REFRESH_GRANULARITY = 1 day`, UTC-day quantization, and one matching day of read/sweep grace. The user then explicitly selected that scheme.
- The former 1-hour minimum caveat is superseded by the final contract in this PR: zero or a whole-day multiple, with one day as the minimum.

Keeping seconds internally is intentional. Dump retention already uses seconds, control-plane transfer already exposes `*_retention_seconds`, and stacked #254 gives Responses and dumps one shared expiration driver with two domain adapters. Only the Responses validator and dashboard constrain the value to whole days.

## Final lifetime semantics

- `responses_retention_seconds = 0` disables durable reads and writes. A disabled HTTP request constructs no durable backing and makes no state-table query.
- Positive values are 1–3,650 whole days, transported and stored as seconds. The dashboard accepts 7 days, 30 days, or a custom integer day count.
- Item and snapshot `refreshed_at` values are floored to the start of their UTC day.
- Read/current-policy/sweep cutoffs all subtract the configured duration plus one fixed day. A quantized row therefore never expires before an exact last access would have expired, and can remain for at most one additional day.
- A same-day item refresh is filtered in the layered store and guarded by `refreshed_at < current_day_bucket` in SQL. The SQL regression checks `total_changes()`: same-day item plus snapshot reuse produces zero mutations; the next day produces exactly two.
- HTTP `store:false` may read earlier durable state when the key opted in, but never inserts or refreshes it. Codex's WebSocket `store:false` session-local state is independent.

There is no visibility floor, epoch, or hidden retention namespace. A later grow may reveal a surviving row inside the wider current window; disable hides all durable state and blocks statement-time writes.

## Concurrency review

Three independent cleanup passes found and drove fixes for:

- a request crossing UTC midnight being incorrectly suppressed by the old request-local “fresh” set;
- a same-day no-UPDATE path accidentally bypassing current key disable/soft-delete validation;
- an item hidden by the request's old short cutoff but made live by a concurrent retention grow being preserved without renewal;
- current-policy growth preserving a conflicting producer ID while exact payload reuse refreshes its lifetime;
- SQL/Memory divergence across those grow, disable, and midnight cases;
- the non-obvious extra cutoff day lacking an adjacent rationale.

The final implementation determines staleness from persisted day buckets, validates insert results against the current API-key policy even when no UPDATE is needed, and rechecks policy again during a real refresh.

## Migration and PR split

`0065_responses_state.sql` directly creates empty exact-ID item/snapshot tables, enforces whole-day retention values and day-aligned refresh timestamps, and creates the owner-neutral `spilled_files` ledger. It carries no compatibility read path.

#249 contains correctness and straightforward cleanup only. Due queues, bounded owner discovery, dump adoption, orphan inventory, and the shared optimized sweep driver remain in stacked #254.

## Verification

Final HEAD `ecc4dbd24`: 395 test files / 4839 tests passed; lint passed; typecheck passed. Focused regressions cover whole-day validation, UTC quantization, same-day zero mutation, next-day refresh, midnight crossing, grow/shrink/disable races, and SQL/Memory parity.

PR #254, opened 2026-07-23
perf(storage): unify expiration sweeps

Menci:
## Summary

Unify Stateful Responses and dump expiration behind one bounded, fair driver.

- Add one `expiration_sweeps(domain, key_id, ...)` due queue with stale-claim recovery, revision-safe completion, fair retries, and a four-unit tick budget.
- Dispatch through exactly two dataset adapters. Responses deletes expired snapshots/items; dumps deletes expired records. Scheduling, claims, fairness, retries, and completion are shared.
- Backfill cleanup tracking incrementally from existing Responses items, Responses snapshots, and dump records with one shared 500-row budget. Each source row schedules its due key; dump rows also register the exact files they still reference.
- Schedule Responses cleanup from each row's UTC-day refresh timestamp, whole-day retention, and one-day rounding allowance; dump cleanup retains exact-second timing.
- Apply each API key's current rolling retention to dump list/get reads, so policy edits change visibility without waiting for physical cleanup.
- Move dump bodies into the shared `spilled_files` registry: per-write unique paths, staged writes, atomic adoption, retirement when the referencing row is deleted, and one shared registered-key collector.
- Remove prefix listing and deletion from `FileProvider`; persistent runtime cleanup is driven solely by stored rows and registered file keys.
- Expand `CHANGELOG.md` into Deployment Notes with `hard`, `minor`, and non-blocking `advisory` levels. Recommended actions may appear at any level and remain explicitly authorized post-deploy work.

## Dependency

This PR builds on the Stateful Responses schema and shared `spilled_files` registry merged in #249. Migration `0066_expiration_sweeps.sql`, dump adoption, bounded cleanup-tracking backfill, and the shared sweep driver are introduced here.

## Migration

`0066_expiration_sweeps.sql` is additive. It creates the `expiration_sweeps` due queue, the `cleanup_backfills` table with three bounded source-table cursors, claim indexes, and scheduling and file-registration triggers. It seeds only the three constant-size cursors: it neither scans source tables nor seeds every API key during migration. Domain/source values are open strings rather than schema enums.

During a rolling deploy, the deterministic dump writer from the preceding Worker version remains accepted only for the exact textual path derivable from its dump row. New per-write unique paths must already be staged. Existing dump descriptors are registered incrementally while their referencing row still exists; deleted rows cannot be resurrected by a stale backfill page.

## Deployment notes

A non-blocking advisory recommends a one-time post-deploy audit of historical `dumps/v1/` objects. Only exact keys absent from both current dump descriptors and `spilled_files` are candidates for manual deletion; runtime code never scans the prefix.

## Risk

The highest-risk surfaces are D1 statement/mutation bounds, claim/completion races, the Responses one-day rounding allowance, hot-key fairness, policy shrink/grow behavior, deployment-overlap writers, and dump descriptor registration. Tests cover those cases, indexed query plans, and adapter statement bounds.

## Test Plan

- [x] Focused expiration, Responses state, dump lifecycle, migration, import, and scheduled-maintenance suites
- [x] `pnpm run test` — 396 files / 4842 tests passed
- [x] `pnpm run lint`
- [x] `pnpm run typecheck`


Menci:
Codex agent — final research, design, implementation, and review record for the shared expiration sweep.

## Why this is a separate stacked PR

#249 contains one coherent breaking change: the Stateful Responses schema cutover and correctness surface (opt-in rolling visibility, statement-time write/snapshot policy checks, payload-identity-safe refreshes, Responses ownership in `spilled_files`, and logically correct cleanup). This stacked PR contains the sweep optimization and dump migration, so the base PR does not introduce a transient optimization architecture that this PR immediately removes.

## Production D1 evidence and resulting bounds

Session archaeology found production `D1_ERROR: D1 DB is overloaded. Requests queued for too long.` around hourly Responses cleanup. The prior shape could cause roughly 234k–293k row mutations in one tick; a 1,413-item response had produced more than 200 SQL statements; D1 batch statements share one execution budget; and an earlier large delete/drop migration failed after 7009 of 7500 operations. These observations ruled out migration-time owner scans/deletes, per-item SQL loops, and unbounded per-key cron work.

The final work budget is explicit:

- cleanup-queue backfill shares 500 rows per tick fairly across three incomplete source tables: Responses items, Responses snapshots, and dump records;
- the expiration driver claims at most four cross-dataset queued keys per tick;
- an adapter deletes at most 100 rows from each source table it handles, using indexed oldest-row probes;
- a partial or failed unit receives a short retry and yields the rest of that tick, so one hot key cannot monopolize the loop;
- the physical collector claims and exact-deletes at most 1,000 keys;
- Responses insert chunks use at most 100 binds, while payload-identity refresh chunks use 94.

`EXPLAIN QUERY PLAN` regressions pin the queue, Responses, and dump range indexes. Active and disabled retention use separate statement-time branches, so active cleanup scans the exact `(key_id, refreshed_at|created_at)` expired range instead of filtering retained history through a broad OR.

## One retention-cleanup algorithm

`expiration_sweeps(domain, key_id, due_at, revision, claim...)` is the only retention-cleanup queue, and `sweepExpirations()` is the only driver. It owns ordering, stale-claim recovery, unit budgets, fairness, retries, and revision-safe completion. Its two adapters contain only domain facts:

- Responses deletes expired snapshots/items and probes the oldest `refreshed_at`.
- Dumps deletes expired records and probes the oldest `created_at`.

There is no all-key migration seed. Three bounded rowid cursors discover API keys already represented in the source tables. Cursor progress is monotonic under overlapping ticks; every discovered key is forced due when necessary, and a new stored row also enqueues its API key atomically. Memory and SQL repositories follow the same queue semantics.

`collectSpilledFiles()` is deliberately separate because it is physical-resource GC, not another retention algorithm. Both adapters only retire registered file keys in `spilled_files`; one shared cleanup path performs all physical deletion.

## Shared dump-file lifecycle

New request/response bodies use independent per-write unique paths. Each path is staged before file IO, adopted atomically by the dump-row insert, retired by owner deletion, and later collected by exact key. A losing duplicate write leaves only its own uniquely keyed staged files collectible; it cannot damage the winning row.

During a rolling `db:migrate:remote && deploy`, the preceding Worker may still write its deterministic descriptor. The migration accepts only the exact textual path derivable from `(key_id, created_at hour, record id, req|resp)`; arbitrary, missing, or null keys fail. Claimed registry entries cannot be adopted. New per-write unique paths always require a staged entry.

Dump backfill registers an existing descriptor only while the matching current dump row still references it, preventing a stale batch from resurrecting a deleted file. Runtime cleanup never lists a file prefix: owner rows retire registered file keys, and shared cleanup deletes only claimed keys.

## Rolling visibility and cursor behavior

No visibility floor or state epoch exists. Responses and dump reads use the API key's current rolling window. Shrinking retention hides rows immediately; growing it may reveal surviving rows that physical cleanup has not removed. Dump pagination anchors the cursor only to a currently visible row; an expired or deleted cursor advances without repeating the first page.

## Review convergence

Independent database/concurrency, architecture, and test-quality reviews drove fixes for:

- later stored rows racing queue completion and backfill scheduling;
- hot-key starvation and due-time loss across revision conflicts;
- retry eligibility accidentally selecting already-claimed work;
- request-cutoff writes racing retention shrink/grow/disable;
- key-only refresh racing an expired-ID payload replacement;
- terminal snapshot writes after disable;
- Memory/SQL missing-key, deleted-key, and stale-claim divergence;
- canonical dump descriptors bypassing adoption checks;
- migration-window old/new writer conflicts;
- empty-key databases being stranded without a queue row;
- dump-row backfill resurrecting a deleted record;
- deleted/expired dump cursors repeating the first page;
- full-key scans hidden behind bounded DELETE limits;
- missing claim-token uniqueness indexes;
- Responses spill hydration selecting a stale column shape.

After #249 merged, the integration pass preserved UTC-day refresh quantization across Memory and SQL while moving every Responses queue due time to `refreshed_day + retention + one-day rounding allowance + 1ms`. This keeps the optimized scheduler aligned with the visibility/deletion cutoffs instead of claiming each key one day early.

A later Claude review correctly identified that prefix inventory was a one-time migration concern embedded as permanent schema and runtime machinery. The final design removes `spilled_file_inventories`, its claim/revision APIs, `FileProvider.listPage`, R2 LIST, filesystem traversal, and the scheduled inventory branch instead of merely adding a completion flag. Historical untracked files are an operational cleanup concern captured as a non-blocking Deployment Note; steady-state writers stage every file in the registry before IO.

All reviewed correctness findings are resolved in the final design. Verification on commit `bf11f4088`: 396 test files / 4842 tests passed; lint passed; typecheck passed.

PR #273, opened 2026-07-27
fix(responses): project collaboration through a plaintext shim

Menci:
Codex exposes a plaintext collaboration dispatch path to non-OpenAI providers, but Copilot reserves the exact `collaboration` namespace and requires its encrypted schema. Forwarding that client contract unchanged therefore makes Copilot produce opaque arguments that another upstream cannot consume.

This PR makes collaboration a reversible Responses source-protocol membrane:

- The client continues to see the ordinary Codex contract: namespace `collaboration`, encrypted message schemas, and `encrypted_function_args: []` on plaintext message actions.
- The selected upstream sees a collision-free ordinary namespace (`collaboration_2`, then `_3`, …) whose message schemas omit `encrypted: true`.
- Function calls, histories, deferred tool inventories, snapshots, and namespace-qualified tool choices are projected in both directions.
- History-only collaboration calls engage the same membrane, including Codex replay that omits the empty plaintext marker for providers not named exactly `OpenAI`.
- Native Responses and representable Messages targets are supported. Chat Completions and Messages shapes that cannot preserve the namespace contract are removed during candidate evaluation.

The collaboration shim wraps the complete server-tool loop, so every internal turn stays in one plaintext alias domain and the client stream is restored exactly once. Context mutation remains one-way, matching the interceptor contract.

Collaboration arguments and `agent_message` content are no longer affinity carriers. The gateway does not inspect or patch inter-agent ciphertext and does not require a subagent turn to land on the same upstream as its parent. Existing structured encrypted tool-output and reasoning carriers remain scoped to their native Responses contracts.

The Copilot item-id membrane likewise treats plaintext agent messages as public items and no longer embeds provider-private affinity into their content.

## Evidence

- Live Copilot and GHE probes reject removing `encrypted: true` from the reserved `collaboration` schema, while ordinary `collaboration_2` / `collaboration_3` namespaces accept the same plaintext arguments and replay.
- Codex dispatches a collaboration message locally when the returned call uses namespace `collaboration` and an explicit empty `encrypted_function_args` list.
- Codex removes that empty marker from replay for providers whose configured name is not exactly `OpenAI`; the ingress projection accepts this plaintext replay form while rejecting explicit null or non-empty encrypted markers.

The previous hosted `multi_agent` implementation on this PR was useful design archaeology for action vocabulary and collision handling, but it was not the protocol authority. Its scheduler, checkpoint format, and `response.inject` server implementation are not part of the resulting design.

## Verification

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] 480 focused collaboration, server-tool, affinity, translation, HTTP, serve, and Copilot item-id tests
- [x] Agent Setup installer verifier: 106 passed, 4 skipped
- [ ] Full workspace: 5405 passed, 5 unrelated web tests fail because the test runtime exposes a malformed `window.localStorage` (`getItem is not a function`)


Menci:
Holding this in draft: nothing has exercised it yet.

The audit that shaped this PR established that **outside the `openai` org nobody has implemented the server side** — every third-party project is either a client or a gateway that forwards to a real OpenAI upstream, and no server anywhere accepts `response.inject`. So there is no reference implementation to compare behaviour against, only the guide and the two reference SDK clients' source and tests.

Everything here is therefore derived from the published contract rather than from observed behaviour, and it has never been driven end to end by a real client. That is the gap to close before this ships: point `agents-python` or `agents-js`'s experimental hosted multi-agent model at Floway and run a real workflow through it. GPT-5.6 is reachable through both Copilot and Azure, so the model is not the blocker — the upstreams simply do not serve the hosted capability, which is the premise for hosting it here.

The contract-derived parts are the ones worth the most scrutiny in review: the per-action argument schemas (only `spawn_agent`'s are documented; the other five are transcribed from Codex CLI collaboration traffic, a different feature reusing the same vocabulary), and the eager serial scheduler, which deliberately diverges from the upstream's asynchronous one because there is no coordinator to run subagents in.

Menci:
I am Claude Code (Opus 5), running for the maintainer. Flagging a defect found while
verifying #456 — the reachability argument there pointed at this branch, and the pointer was
correct even though the conclusion drawn from it was not.

`opaque-locations.ts` at `e1d0911` is a net **narrowing** of the affinity carrier set. It
gains the `function_call_output` / `custom_tool_call_output` `output.N.encrypted_content`
slots, and it loses `agent_message`'s `content.N.encrypted_content` on every side at once:

- `responsesOpaqueLocations` has no `agent_message` branch, and an `agent_message` has no
  top-level `encrypted_content`, so it yields zero locations for those items;
- the egress synthetic-carrier branch that appended
  `content.${item.content.length}.encrypted_content` is gone;
- `canCarryAffinity` drops `'agent_message'` from its list;
- the ingress scan loop that walked `item.content` is gone.

`ResponsesAgentMessageContent` still carries the `encrypted_content` variant (now spelled
`ResponsesEncryptedContent`), so this is not a slot that stopped existing — it is a slot we
stopped looking at.

Two concrete consequences, both on carriers that are already sitting in client conversation
history from builds before this branch:

1. **A synthetic carrier reaches a third-party upstream.** On `main`, `ensureItemCarrier`
   appends an `encrypted_content` content part to an `agent_message` whose payload is 100%
   Floway ciphertext with zero upstream-origin bytes, and ingress removes it before dispatch
   (`projectOptionalAffinityBlob` returns `remove` because `decoded.value === undefined`).
   After this branch, ingress never decodes it, so the removal never happens and we transmit
   our own ciphertext to the provider.
2. **An ordinary carrier reaches the upstream with our trailer attached.** Same mechanism,
   for an `agent_message` content part that wraps real upstream bytes: unscanned, so
   unwrapped, so forwarded framed.

Both are exactly the impact #456 describes, arriving through our own change rather than
through the operational events that issue hypothesized. The affinity pin carried by those
items is lost along the way.

The narrowing is the bug, and it is the second time this file's slot inventory has been
restated by hand. Extracting `responsesOpaqueLocations` was the right move; the fix is to
finish it so the inventory cannot diverge again:

- restore the `agent_message` `content.N.encrypted_content` locations, including the
  synthetic-carrier append path, so egress and ingress both derive from the single inventory
  with no per-side item-type lists;
- reduce `canCarryAffinity`'s residual literal list to something the inventory itself
  determines, so adding or removing a carrier slot is a one-place edit;
- pin every item type × slot with a wrap-then-unwrap round trip, so a future narrowing fails
  the build rather than silently stranding carriers in client history.

I am picking this up now.

Menci:
Fixed in 9ff9714d1, pushed to this branch.

The inventory now enumerates each slot once as a path, and reading, rewriting, growing and removing a carrier all derive from that single enumeration — `canCarryAffinity` and `ensureItemCarrier` no longer keep their own item-type lists, and a tool output stays a slot we may rewrite but never grow.

Worth recording: the existing `agent-message natural and originless nested carriers round-trip` test passed against the narrowing. It asserted only that the value survives egress and ingress, which is trivially true when nothing wraps it. The new table-driven test asserts egress *changed* the value and that no wrapped value ever reaches a mismatched upstream, walking every item type and slot. Reverting the fix fails it on `agent_message` for exactly that reason.

Verified: lint clean, typecheck clean, 5485 tests pass.

Menci:
I am Codex. I ran live collaboration experiments for `deepseek-v4.1-flash` and `gpt-5.6-luna`, then repeated both against the synchronized PR head before squash.

## Environment and method

- Codex CLI **0.155.1** → a local Node Floway instance → a remote Floway configured as the sole Custom upstream.
- Initial code: `1c2702603877f9770ccbdc10e2b8bb24b85660ad`. Post-sync verification: `a1381cf9ae10cd5c896606ea98c9cddeb4d6b132`.
- Separate temporary `CODEX_HOME` and workspace for each model; HTTP Responses/SSE, `supports_websockets = false`, low reasoning effort, and `multi_agent` / `multi_agent_v2` enabled.
- Each home used the local Floway Codex catalog through `model_catalog_json`. Both entries already advertised `multi_agent_version: "v2"`; the experiments required **no catalog or other production code changes**.
- The collaboration shim used the normal provider default. The local instance had its own database, API key and request capture, isolated from the remote gateway's configuration.
- Each main agent created exactly one same-model subagent with `fork_turns: "none"`, instructed it to acknowledge a nonce and remember **23**, then sent a `followup_task` asking that same agent to add **19**. Further agent recursion was disallowed.
- Assertions inspect the persisted **parent and child session JSONL**. CLI stdout and a parent's claim of success alone were not accepted as proof.

## Results

| Run | Main and child model | Collaboration calls | Plaintext agent messages | Exact child replies |
|---|---|---:|---:|---|
| Initial | `deepseek-v4.1-flash` | 3 | 4 | `INITIAL_ACK_DEEPSEEK_7391`; `FOLLOWUP_ACK_DEEPSEEK_7391 42` |
| Initial | `gpt-5.6-luna` | 4 | 5 | `INITIAL_ACK_LUNA_7391`; `FOLLOWUP_ACK_LUNA_7391 42` |
| Post-sync | `deepseek-v4.1-flash` | 4 | 4 | `INITIAL_ACK_DEEPSEEK_7391`; `FOLLOWUP_ACK_DEEPSEEK_7391 42` |
| Post-sync | `gpt-5.6-luna` | 4 | 4 | `INITIAL_ACK_LUNA_7391`; `FOLLOWUP_ACK_LUNA_7391 42` |

Both post-sync runs performed the exact sequence `collaboration.spawn_agent → collaboration.wait_agent → collaboration.followup_task → collaboration.wait_agent`.

The initial Luna worker additionally called `collaboration.send_message` to send an early acknowledgment. Its parent queued the follow-up after that message, before the exact initial final answer arrived. The exact initial answer and follow-up result nevertheless appear in both worker output and parent-delivered messages. This is documented in the archive rather than counting the parent's summary as the exact initial reply.

The initial run made **17 local gateway requests**; the post-sync run made **14**. All **31** returned HTTP **200**, with no recorded request errors.

## What the JSONL proves

- Parent and child `turn_context` entries name the intended model and `multi_agent_version: "v2"`.
- Raw `response_item.function_call` entries carry `namespace: "collaboration"`, proving the V2 collaboration path.
- Across both runs, all **9 message-bearing calls** (`spawn_agent`, `followup_task`, and the initial Luna `send_message`) contain readable message arguments and explicit `encrypted_function_args: []`.
- All **17 `agent_message` entries** contain only `input_text` parts: **zero `encrypted_content` parts** in inter-agent messages.
- Child session metadata links each child to the correct parent at depth 1. No grandchildren were spawned.
- Each child received the initial task and follow-up as plaintext, emitted both exact answers, and those answers were delivered back to its parent. The second prompt did not repeat 23; the correct result 42 verifies retained child context.

Post-sync line-number index (the archive preserves JSONL entry order and line numbers):

| Evidence | `deepseek-v4.1-flash` | `gpt-5.6-luna` |
|---|---|---|
| Parent session | `01a0caa6-4402-7c11-9587-89200ebe815b` | `01a0caa6-43f9-7323-ae21-5feaa8926e96` |
| Child session | `01a0caa6-68f9-7a40-9a9b-205fbd32dd98` | `01a0caa6-5e69-7483-89da-b8c7677ce1bc` |
| Parent V2 context | 8 | 8 |
| Parent spawn / follow-up | 15 / 35 | 13 / 31 |
| Child plaintext incoming tasks | 10 / 21 | 10 / 21 |
| Child exact replies | 14 / 25 | 14 / 25 |
| Parent-delivered exact replies | 30 / 49 | 28 / 45 |

These results cover the two tested model routes and the recorded commits. They do not establish universal upstream behavior or Stateful Responses continuation support. Any opaque reasoning blobs in a session are outside the inter-agent plaintext assertion.

## Reproducible evidence attachment

[Download the complete sanitized experiment evidence](https://github.com/user-attachments/assets/2970b763-e607-42c8-9721-e1394c85989b)

The attachment is a ZIP archive stored with a `.zip.png` carrier filename because the attachment API accepts that upload type. Save it as ZIP and extract it:

```sh
curl -fsSL https://github.com/user-attachments/assets/2970b763-e607-42c8-9721-e1394c85989b -o collaboration-live-evidence.zip
unzip collaboration-live-evidence.zip -d collaboration-live-evidence
cd collaboration-live-evidence
python3 verify.py
```

Archive SHA-256: `9d8183ab4cb0667cb963bbb63abb536afaf4bad7ea3b03816a19a00de66df8fb`.

The archive includes all eight parent/child session JSONL files, both runs' indexed assertions and request metadata, prompts, CLI output, final replies, selected catalog entries, sanitized effective configuration, an offline assertion script and per-file checksums. Model routing labels, local paths and the private remote hostname are normalized; credentials and runtime databases are excluded. Original and sanitized JSONL checksums are recorded separately. Offline assertions pass for all four cases.

PR #275, opened 2026-07-28
refactor: price from the upstream's own usage instead of the client's

Menci:
Pricing read the usage Floway sends the **client** — a wire projection whose protocol may have no field for a bucket we are billed for. An Anthropic 1-hour cache write reaching a Responses client has nowhere to go; Gemini has no cache-write count or service tier at all. So a `Symbol('usage-billing')` key smuggled the missing facts alongside the client's own usage object.

The symbol was chosen for a real property — it survives spread, and `JSON.stringify` drops it — but `Object.entries` and `structuredClone` drop it too, and both had already cost us: gemini affinity egress carried an explicit re-attach with a defensive `throw`, and the Responses server-tool shim's usage merge dropped it silently, mispricing every multi-turn shim response.

A better hiding place was not the fix. Pricing now reads what we bought.

## BillableUsage

```ts
export interface BillableUsage {
  input: number;        // exclusive of cacheRead and both cache-write buckets
  cacheRead: number;
  cacheWrite: number;
  cacheWrite1h: number;
  output: number;
  tier?: string;
}
```

Complete rather than residual: it is the **sole** input to pricing, and the client-facing usage is never read. Exclusive/split form, because that is how the buckets are priced rather than how any one protocol reports totals. Nothing is carried that pricing has no dimension for.

It is produced once, at `providerStreamResultToExecuteResult` — the single choke point every protocol's upstream call funnels through, which already observed each upstream frame for first-output-token timing. Cross-protocol targets recurse into the other protocol's attempt, so the layer that calls the provider always matches the upstream protocol, and the extractor is chosen there.

Delivered on `EventResultMetadata.billableUsage`, beside `modelIdentity` and `performance` — the existing gateway-internal, never-serialized, per-turn channel. Not a per-frame field and not a frame variant: only the terminal event ever reports usage.

A response can span several upstream turns — the server-tool shim's ReAct loop — and we are billed for each, so the shim sums them onto the metadata it publishes. Because every streaming response now resolves this promise and the respond layer awaits it in a `finally`, an abort settles it too, so a transport that abandons the stream cannot hang that await.

## Falls out of it

- **`packages/translate` knows nothing about billing again.** No reads, no writes, no threading.
- **An inclusive/exclusive round trip on the money path disappears.** A Messages upstream reports `input_tokens=10` exclusive; the translator converted it to `45` inclusive for a Responses client; pricing split it back to `10`. Pricing reads the `10`.
- **`splitCacheWriteTokens` no longer takes a billing argument** — an upstream that splits cache creation by TTL reports the split natively; only the translated view ever lost it.
- **`SourceStreamState.usage` and its per-frame accumulation are gone**, along with the `hasTokenUsage` guard and the `observeUsage` switch that selected them. They existed to build a figure from the frames heading to the client.
- **The `cloneGeminiEvent` workaround is deleted.**

Gemini has no native upstream target, so nothing ever reads a Gemini upstream's usage; a Gemini client is priced from whichever protocol served it.

Net: **383 insertions, 1345 deletions.**

## Notes for review

Tests that asserted symbol-specific behaviour are gone rather than adapted — `reassembleGeminiEvents` preserving it, affinity egress preserving it across the clone, the Messages streaming test that derived usage from observed client-facing frames, and the `SourceStreamState` usage-tracking suite. All covered mechanisms that no longer exist. The four usage suites now target the upstream-protocol extractors directly, and there are new regressions for the shim's summing and for the abandoned-stream settle.

## Test Plan

- [x] `pnpm run typecheck` — clean across all 21 workspace projects
- [x] `pnpm run lint` — clean
- [x] `pnpm run test` — 420 files / 4893 tests passing


Menci:
Superseded design — the envelope field on `EventFrame` is wrong and this branch will be rewritten before review.

Objection that killed it: `EventFrame` is the universal envelope for every streaming event — every text delta, every item lifecycle event. An optional `billing` on it claims any event may carry billing, when only the terminal one ever does. That is the same category error as the symbol, one level up.

## Target design

**`BillableUsage` replaces `UsageBillingMetadata`.** It is not metadata and not residue — it is the sole authority for pricing. The name says what it is.

```ts
export interface BillableUsage {
  input: number;        // exclusive: excludes cacheRead and cacheWrite
  cacheRead: number;
  cacheWrite: number;   // 5-minute bucket
  cacheWrite1h: number;
  output: number;
  tier?: string;
}
```

Canonical exclusive/split shape, camelCase like the other gateway-internal metadata. Maps 1:1 onto the gateway's `TokenUsage`.

**Two invariants:**

1. `BillableUsage` carries *everything* pricing needs. Pricing reads no other field.
2. Pricing never touches client-facing data.

**Consequences.** The inclusive/exclusive round trip disappears. Today a Messages upstream reports `input_tokens=10` exclusive, the translator converts it to `45` inclusive for a Responses client, and pricing splits it back to `10` — a lossy-looking conversion undone on the money path, purely because pricing reads the translated object. Under these invariants pricing reads the upstream figure directly.

Protocol expressibility gaps stop mattering for billing: a client whose protocol has no 1-hour cache-write field still cannot see that split, but nothing needs to smuggle it, because pricing no longer reads the client's object.

**Channel.** `EventResultMetadata.billableUsage`, alongside `modelIdentity` and `performance` — the existing gateway-internal, never-serialized, per-turn metadata channel. Not a per-frame field, not a frame variant.

**Population.** Extract from the upstream result in the upstream's own protocol, so no translator participates. The four existing `tokenUsageFrom*Result` bodies are already the right logic; they move into `packages/protocols` returning `BillableUsage`, and run against the upstream object rather than the client one. Same-protocol passthrough is then free — the two objects are the same. `packages/translate` reverts to knowing nothing about billing.

**Deletions.** `USAGE_BILLING`; the `billing` field on `EventFrame`; every translator read/write site; `splitCacheWriteTokens`'s billing parameter; the `cloneGeminiEvent` structuredClone workaround.

## Why the symbol had to go

It was chosen for a real property — survives spread, dropped by `JSON.stringify` — but `Object.entries` and `structuredClone` drop it too, and both already cost us: an explicit re-attach with a defensive throw in gemini affinity egress, and a silent drop in the Responses server-tool shim's usage merge that mispriced every multi-turn shim response.

PR #277, opened 2026-07-28
feat(web): rebuild the dashboard as a React SPA restyled onto WinUI 3

Menci:
## Summary

Rebuilds the dashboard as a React Router SPA on Vite, restyled onto WinUI 3, and makes the control-plane contracts it reads gateway-owned.

- **The app.** React Router in framework mode with runtime server rendering off; client loaders act as resource barriers, so a route's URL and component tree commit only once its resources have resolved and there is no content-level loading state after boot. The document never scrolls: every scrollable region declares its axes through one `ScrollArea`, which enables OverlayScrollbars only where native scrollbars consume layout.
- **A WinUI 3 layer over Fluent.** `apps/web/src/winui/` transcribes the WinUI theme dictionaries into `--winui-*` custom properties, re-points the Fluent tokens with a direct counterpart, and restates per control what a token substitution cannot say. Every value carries a permalink into `microsoft-ui-xaml`, and a departure is written down at the rule that departs. Fluent resolves `appearance`, `size` and `intent` in JavaScript and writes nothing a selector can name, so the layer stamps the resolved value back onto the DOM as `data-winui-*`.
- **Motion from the platform.** Durations and easings are transcribed rather than chosen — dialog, drawer, expander, page transition, toast, and the reposition of a stack — including values that exist only in the OS visual style's PVL table and are recovered through WinJS's transcription.
- **Every surface.** Upstream and model editing with lazy Monaco YAML, aliases, proxies, API keys with Agent Setup, request dumps over SSE, Playground, Usage, Performance, Users, Search, backup/restore, and API Docs generated from the same method/path manifest gateway registration consumes.
- **Localized** in English and Simplified Chinese, with a parity suite over the two tables and one locale fetched per session.
- **Gateway contracts** made projection-specific: upstream records, usage views, dump pagination, API keys, backup validation, and generation-safe sessions.

Navigation is addressable: anything whose click lands on a URL worth opening beside the current one — a record, a page, a time range, a filter — is a real anchor, so middle-click and ctrl-click work. A click that moves within one editing session stays a button.

## Test plan

- [x] `pnpm run lint`
- [x] `pnpm run typecheck`
- [x] `pnpm run test`
- [x] `pnpm run test:agent-setup-installers`
- [x] `pnpm --filter @floway-dev/agent-setup run generate-assets --check`
- [x] `pnpm run build:web`, including the Monaco-lazy, gallery-absent and locale-split assertions
- [x] headless Chrome at 1600×1000 and 390×844: every dashboard route reached signed-in, with no error boundary, no Vite overlay, no document scrolling in either axis, and no uncaught page error
- [x] headless interaction checks: the row Edit action carries an upstream address, the usage range items are anchors carrying their full serialized state, the request list and API Docs render their content


Menci:
Notes from the work that have no home in the code — negative results, deliberate non-changes, and what is still open. Recorded here so none of it has to be re-derived.

## Checked and found false

Each of these was believed at some point in this branch's history and disproved against source. Several were believed by more than one reviewer.

**Dependencies**
- `isbot` is not an unused dependency. `@react-router/dev`'s entry resolution rewrites `package.json` and shells out to a package-manager install whenever an app has no `entry.server` file and the dependency is absent. Removing it makes `react-router typegen` — which this package's typecheck and build both run — mutate `package.json` and the lockfile mid-run.
- The `dompurify` override is not a no-op. `monaco-editor` 0.54.0 pins it at exactly 3.1.7, so the override is a security floor, not leftover pinning (GHSA-c2j3-45gr-mqc4, fixed in 3.4.12).

**WinUI**
- **WinUI 3 ships no toast control.** InfoBar is inline, persistent, and has *zero* animation — its show/hide is a bare `Visibility` swap. TeachingTip is transient but has no severity concept, no `Duration`/`Timeout` property, and positions bottom-centre. ContentDialog is modal and throws `ERROR_CONTENTDIALOG_MULTIPLE_OPEN` on a second open.
- `StackedNotificationsBehavior` **does not stack**, despite the name. It is a strict FIFO queue showing one notification at a time, persistent by default (`Duration` defaults to `null`), and the maintainers rejected stacking by name: *"realized it was a more disruptive user experience in general to replace an active notification"*.
- **No Microsoft surface shows a countdown for a self-dismissing notification.** The evidence is absence plus completeness: the toast XML schema is a closed enumerated element set with no timer element, and there is no API to query remaining time. Microsoft's answer to "the user might miss it" is a system-wide duration setting, not a per-toast indicator.
- `CommunityToolkit.WinUI.Notifications` renders nothing — 132 C# files, zero `.xaml`, zero references to any XAML base type. It builds OS notification payloads, and is deprecated in favour of the Windows App SDK.
- TeachingTip does **not** "grow from 0.01 scale": the keyframe is `Min(0.01, 20.0 / Width)`, so 0.01 is a *ceiling* and the intent is "start as a 20×20 nub". Its motion is also not in `TeachingTip_themeresources.xaml` at all — it is hand-built Composition in the `.cpp`, so anyone following the older note looks in the wrong file.
- `AddDeleteThemeTransition`'s unload leg **does** translate (`sourceoffset = start - destination`); an earlier claim that both legs are pinned to `{0,0}` cited two functions that do not exist at the pinned SHA. What *is* true is that `PopOutThemeAnimation` hardcodes both offsets and exposes no offset property — that single authority is what the toast's exit rests on.

**Web platform / libraries**
- `role="radio"` on `<a href>` is **explicitly conformant** — ARIA in HTML lists `radio` among the allowed roles for an anchor with an href. APG's own toolbar example uses radio semantics for a segmented control built from buttons, on the stated principle that ARIA conveys *behaviour*, not element type or styling.
- **Radix Primitives is not deprecated.** A seven-month dormancy fed the rumour; there were 21/90/105 commits in May/June/July 2026. What *is* npm-deprecated is `@base-ui-components/react`, renamed to `@base-ui/react`.
- Fluent's toast **can** be used headlessly. Only `renderToastContainer_unstable`'s body is welded shut; `onMotionFinish(ev, { direction })` is publicly typed on `ToastContainerState`, so the state machine does not care what animates it.
- The premise that a headless library supplies behaviour while the app supplies motion is **false for exit motion** almost everywhere: Base UI's Toast has no `actionsRef` hatch, React Aria splices synchronously, and Zag hands you a fixed 200 ms `removeDelay` to mirror by hand. Only Radix's `forceMount` is a genuine hatch — and Radix ships no queue at all.
- The dashboard's own message bar already draws WinUI's disc-plus-knockout severity icon in pure CSS (a `radial-gradient` behind a `*Filled` glyph whose symbol is negative space), which refutes the claim that reproducing it requires replacing the component.
- Commit `755eefa65`'s message is wrong about its own mechanism: it argues the listbox overflow changed on classic-scrollbar platforms, but the predicate runs the other way and the change is a visual no-op on every platform. The code is fine; the message is not.

## Looked at and deliberately kept

So that a later fidelity pass does not re-open them:

- **Cursor policy.** Fluent's hand on interactive surfaces and barred circle on disabled ones are kept as a web-platform departure, recorded in the layer. WinUI states no cursor at all, and the app's own non-Fluent controls already chose the web idiom in eleven places — a half-applied policy would be worse than either whole one.
- **Overlay beaks** kept. Their stroke is already WinUI's `SurfaceStrokeColorFlyout`, and WinUI itself ships a beaked overlay in TeachingTip.
- **Table row hairlines** kept — legible density, not brand identity. The difference between card-hosted tables (no rule under the last row) and standalone ones (rule kept) is deliberate: the card's own border is that line, and a standalone table's last rule is its own bottom edge.
- **The toast's severity glyph** kept, on InfoBar's reading; its colour comes from WinUI's `SystemFill` family.
- **Toast motion** left as Fluent's: the container hard-codes its animation with no motion slot, so matching WinUI means CSS fighting a JS animation.
- **Three spellings of the gateway path set** (bare prefixes, wrangler globs, nginx regexes) stay. They are three languages with different semantics, and `scripts/check-wrangler.ts` compares the set in both directions — regenerating them at one width would make `pnpm run deploy` refuse on every existing checkout until each operator hand-edited their gitignored `wrangler.jsonc`.
- **Forced-colours handling** left as it is; there is no High Contrast palette to hold it to.
- **The 720px width** on the three wide editor dialogs is a house value, above both WinUI's 548 and Fluent's own 600, and now says so.
- **Usage metric tiles** stay `ToggleButton`s rather than links: their whole state layer keys off `aria-pressed`, which is invalid on an anchor.

## Bundle, measured

| | before | after |
|---|---|---|
| Deployed assets | 51.15 MiB | **8.72 MiB** |
| Login page, brotli | 492.8 KiB | **445.1 KiB** |
| Entry chunk, brotli | 27.4 KiB | **4.2 KiB** |

Source maps are still generated — two build checks read them to decide chunk membership — but excluded from the upload via a generated `.assetsignore`, and `sourcemap: 'hidden'` so no deployed chunk carries a dangling `sourceMappingURL`. Merging the small shell chunks was measured and **rejected**: the only workable form saved 0.4 KiB brotli for a hand-maintained path list, and the principled form (`tags: ['$initial']`) swallowed the whole route tree for a 247 KiB regression.

## Still open

- **Segoe UI Variable is the largest single cost on a first visit.** 1,074,054 B of uncompressed TTF from a third-party origin, `rel="preload"`, served without content-encoding and effectively incompressible. It takes a first visit from 445 KiB brotli to **1494 KiB**. Every fix has a cost: self-hosting or subsetting raises licensing; dropping the preload trades the bandwidth for a font swap.
- **Session failure redirects on two pages and errors on the rest.** react-router's `findRedirect` runs before error processing, so a thrown redirect beats a sibling loader's thrown Error. Usage and Performance therefore send you to the login page where every other dashboard page hits the error boundary.
- **The Fluent namespace import defeats tree-shaking**, worth ~71 KiB brotli on every route — source-map attribution names 282.6 KiB raw of components the app never references. A naive conversion to named imports builds, but 500s during prerender on the CJS/ESM interop the current code was written to survive.
- **An interactive Fluent `Card` repaints every descendant `Text`** to its own foreground through a two-class selector no utility can outrank. Nothing in the product hits it — an ESLint rule now forbids the props that make a Card interactive — but the real repair would put Fluent's sheet in a lower cascade layer, which is a change to how every style in this app is injected (researched: feasible, but only by wrapping a `@private` Griffel member).

PR #278, opened 2026-07-28
fix(web): rebuild the API keys Agent Setup panel

Menci:
Stacked on #277. Scope: the API Keys page only — the font work is #280 and the login-page copy is #281.

## Summary

- give the Agent Setup panel a sidebar: connection method as a horizontal tab list on top, agent navigation below, replacing the tab row plus `Connection method` dropdown
- move the platform switch onto the code block header where the language caption sat, and drop the standalone `Platform` field
- render the "select a key" and "preparing" states as a shell comment inside the command block, with Copy disabled, instead of a message bar above the form
- lay every configuration field on one column track; the tracks floor at zero and the wrapped controls zero their native `min-width`, because a combobox otherwise reports its input width as min-content and pushes the column past the panel
- hang field explanations off Fluent's `InfoLabel` info button through a shared `infoLabelSlot` helper
- move the panel title and the key-in-use note into the card: the heading reads "Setup Your Agents" and the key is named in the note under the command
- drop the Codex mark's full-bleed app-icon plate, which rendered its glyph a quarter smaller than every other provider mark at the same box size

## Verification

- `pnpm run lint`
- `pnpm --filter @floway-dev/web run test` — 33 files, 188 tests passed
- manual: Node gateway + Vite, screenshotted Claude Code / Codex × Setup script / Config snippet × both platforms, and measured the panel's layout tracks in a headless Chrome to confirm nothing overflows


Menci:

PR #280, opened 2026-07-28
fix(web): correct the dashboard's stylesheet delivery

Menci:
Stacked on #277.

## Summary

Two defects in how the dashboard's CSS reaches the browser.

**Utility stylesheet generation.** UnoCSS ran in `per-module` mode, which generates one sheet per module and leaves the bundler to concatenate them in module-graph order. That forfeits UnoCSS's rule ordering, which is a property of a single `generate()` call. Breakpoint variants landed ahead of the base utilities they override, and since a media query adds no specificity the base utility won at every width — 42 responsive rules across 25 files were inert, among them the sidebar collapse and the mobile navigation trigger.

`per-module` was itself a workaround: the Vite integration's default global mode emits nothing under React Router. It keys its `vite:css-post` handle by the top-level `build.outDir`, while React Router sets `outDir` only per environment (`dist/client`, `dist/server`) and opts into `builder.sharedConfigBuild`, which is what would otherwise re-resolve the config per environment. The lookup misses and the layer placeholder ships as the entire stylesheet — [unocss/unocss#4990](https://github.com/unocss/unocss/issues/4990), open, and not specific to Vite 8 or Rolldown.

Generating through `@unocss/postcss` sidesteps that machinery entirely: one pass over the content set is one `generate()` call, so base utilities, `lt-*`, and ascending `min-width` breakpoints emit in that order. The plugin's `cwd` is pinned to the package, because it resolves both config discovery and the `content.filesystem` globs against the build process's working directory — without the pin, a build launched from the workspace root emits an empty sheet. The source-level transformers go with the Vite plugin, which was the only thing applying them; neither was reachable, since the source has no variant groups and no CSS files for the directive transformer.

**Monospace face.** Bundle Cascadia Code through Fontsource and publish it as the theme's `fontFamilyMonospace`, so code surfaces stop resolving to Courier on macOS. Route UnoCSS's `font-mono` and every remaining hardcoded `monospace` declaration through that one token, and override the Prism themes' own font declaration, which addresses the code block through `pre[class*="language-"]` and outranks a single-class rule. Drop the `fontFamilyBase` / `fontFamilyNumeric` overrides, which pinned the bare `sans-serif` (Helvetica on macOS) in place of Fluent's Segoe UI stack.

## Verification

- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run test` — 420 files, 4,772 tests passed
- `pnpm --filter @floway-dev/web run build`
- parsed the emitted stylesheet and cross-checked every responsive class against the source: 42 overridden before, 0 after
- confirmed the computed layout at 500, 700, 900, and 1400px in a headless browser, and measured the rendered `font-family` and loaded `@font-face` against the running dashboard


Menci:

PR #281, opened 2026-07-28
fix(web): keep the dev passwordless shortcut off the login page

Menci:
Stacked on #277.

## Summary

- the sign-in hint published that an instance without `ADMIN_KEY` accepts a blank password; that is a property of the deployment, not guidance for a login form
- the hint now states only the seed-admin path: a blank username with `ADMIN_KEY` as the password

## Verification

- `pnpm --filter @floway-dev/web run test` — locale parity suite included


Menci:

PR #282, opened 2026-07-28
fix(web): rebuild the Playground's model surface

Menci:
Stacked on #277.

## Summary

- drop the Generation section's temperature, max output tokens, top P, frequency penalty, presence penalty and stop sequence controls; the Custom JSON editor already expresses those request fields, and reasoning effort — whose accepted values come from the model catalog — stays
- restore the model capability summary the Vue dashboard had: provider badges, advertised token limits, and for an alias what it resolves to under the caller's upstream cap, rendered as Fluent outline badges in the Playground header
- fix the model picker's search, which could not match while a model was selected

## Notes

Badge assembly is a pure catalog walk in `model-badges.ts`, so alias reachability, the in-cap binding union and the "varies" collapse for rules the targets disagree on are covered without rendering. `reachability.ts` gains `reachableTargets`, shared with `isModelReachable`; `availableModels` now takes the resolved cap the header needs anyway.

## Verification

- `pnpm exec vitest run --project @floway-dev/web` — 37 files, 200 tests passed
- `pnpm exec eslint apps/web`
- manual: seeded a local instance with multi-upstream models, single-target / multi-target / partially-reachable / unresolvable-target aliases, and a live Copilot upstream; verified every badge shape and that model search filters


Menci:

PR #283, opened 2026-07-28
fix: name the Microsoft search provider Web IQ

Menci:
Stacked on #277.

## Summary

**Name the provider what it is.** The Microsoft web search provider has always called `api.microsoft.ai/v3/search/web` and `/v3/browse` with an `x-apikey` header. Those are **Microsoft Web IQ** endpoints. The product was unreleased and documented only on Microsoft's intranet when the provider landed, so it shipped under the internal name "Microsoft Grounding" — a name that was never public.

That matters beyond tidiness. The name it most resembles, **Grounding with Bing Search**, is a different service that is still generally available, on `api.bing.microsoft.com` behind an Azure `Microsoft.Bing/accounts` resource. An operator reading the old label would provision that resource and receive a key this provider cannot use; the Web IQ endpoint ignores `Ocp-Apim-Subscription-Key` outright and answers `AuthNoToken`. Microsoft's own FAQ keeps both products alive side by side, so this was a misnomer pointing at a live product, not a stale name.

- rename the provider identifier to `web-iq` across the config contract, the stored column, and recorded usage, migrating existing rows so credentials keep working and per-provider usage history stays continuous
- label it **Microsoft Web IQ**, the name the product uses for itself, and link to the Web IQ portal where its keys are actually issued rather than the retired Bing APIs page
- drop the Bing mark for the Microsoft one — Web IQ publishes no mark of its own, only the Microsoft four-square
- derive the `search-usage` provider rejection message from the provider list; it named two of the three providers and would have drifted again here

**Retire the pre-rename export format.** A backup written before this change spells the provider key `microsoftGrounding`, which strict parsing rejects with a field-level complaint about a file that is simply too old. Bumping the envelope to 18 makes it fail on the version with an actionable message, and keeps compatibility shims out of the parser. The dashboard kept its own copy of that version, which the bump would have left one behind — enough to reject every backup file the deployment writes — so it is now annotated with the gateway's own literal and the next bump breaks the build instead of failing at runtime.

**Finish the Fluent migration on this page.** Driving it in a browser against a live Web IQ rejection surfaced four rough edges:

- the OK/Error pill and the failure box carried hardcoded `light-dark()` hex pairs; they are a Fluent `Badge` and a `MessageBar` now, the same MessageBar the save path already uses
- the upstream and model pickers had become bare native selects, dropping the per-option descriptions the previous dashboard showed; they are Fluent `Dropdown`s again, an upstream carrying the same kind description the upstreams list uses and a model carrying its id only when it differs from the display name
- the summary line printed the raw provider identifier with hardcoded English around it; it reads the provider's label now, falls back to the id verbatim for a provider this build does not know, and goes through i18n
- the provider error extractor had no rung for Web IQ's error envelope, so its whole JSON body reached the operator and overflowed the panel; reading `userMessage` turns that into the one sentence Microsoft wrote for it

## References

- https://webiq.microsoft.ai/
- https://github.com/Azure/azure-rest-api-specs/pull/43848 — "AIGrounding -> WebIQ rebrand"

## Verification

- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run test` — 423 files, 4,754 tests passed
- manual: drove the page in a headless browser against a Node gateway — provider picker, both passthrough pickers with their descriptions, and a live `AuthInvalidApiKey` rejection from Web IQ


Menci:

PR #284, opened 2026-07-28
fix(web): floor every dashboard page region at zero width

Menci:
Stacked on #277.

## Summary

- every route wraps its content in a single-column grid marked `min-w-0`, which lets the region shrink but leaves its column at `auto`; an auto column grows to the widest child's min-content, so one long message bar pushed the whole page past the viewport and took the page header and its primary action off screen
- the wrapper becomes a `dashboard-page` UnoCSS shortcut that pins the column to `minmax(0, 1fr)`

## Verification

Reproduced on the Upstreams page with the real warning bar and measured `main` before and after:

| | `main.scrollWidth` | `main.clientWidth` | page track |
|---|---|---|---|
| before | 1455 | 1210 | 1427px |
| after | 1210 | 1210 | 1154px |

With the column floored, the message bar switches to its multiline layout and wraps rather than truncating. Wide content inside a panel no longer widens the page either (checked by injecting a 400-character unbreakable string into a panel: `main.scrollWidth` stays at the viewport width).

- `pnpm run lint`
- `pnpm --filter @floway-dev/web run test` — 188 tests passed


Menci:

PR #300, opened 2026-07-30
fix(responses): reject missing models before routing

Menci:
## Summary

- validate Responses request models at the canonical source boundary
- return a Responses-shaped 400 with `param: "model"` for missing or non-string values
- keep malformed compact requests out of alias lookup and model resolution

## Root cause

The wire payload was cast to `ResponsesRequestPayload` without validating `model`. A request that omitted it therefore carried `undefined` into `SqlModelAliasesRepo.getByName`; both node:sqlite and D1 reject that bind value and surfaced a 500.

Current OpenAI OpenAPI and SDK contracts require the compact `model` field, and Codex has always serialized its active model for this endpoint, so this is a client input error rather than a model-defaulting case.

## Verification

- `pnpm run test` — 422 files, 4907 tests passed
- `pnpm run lint`
- `pnpm run typecheck`

### DIFF
--- a/packages/gateway/__tests__/data-plane/chat/responses/http_test.ts
+++ b/packages/gateway/__tests__/data-plane/chat/responses/http_test.ts
@@ -416,6 +416,23 @@ test('POST /v1/responses returns 502 when the response snapshot cannot be persis
   }
 });
 
+test('POST /v1/responses/compact rejects a missing model before resolution', async () => {
+  installRepo();
+  lastSeenModel.value = null;
+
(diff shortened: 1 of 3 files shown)


Menci:
Superseded by #313, which fixes the same crash at the same seam (`canonicalizeResponsesPayload`) and additionally carries the OpenAI-shaped `code: missing_required_parameter`, removes the duplicated WebSocket-side checks, and covers HTTP generate, HTTP compact and WebSocket through the single funnel.

PR #306, opened 2026-07-30
fix(responses): send the spec's `status` key on WebSocket error frames

Menci:
First of four PRs recreating the WebSocket conformance stack against the OpenResponses `2026-04-24` specification. The branches survived an unauthorized merge-and-revert cycle on `main`; the code is unchanged, this restores the review surface.

## Defect

The Responses WebSocket transport serialized its JSON error envelope as `{ type: 'error', status_code, error }`. The spec's `WebSocketErrorEvent` schema pins the required keys to exactly `type`, `status`, and `error`, so `status_code` was a Floway-private key with no other producer and no other consumer.

The practical consequence is narrower than it sounds, and worth stating precisely: Floway already emitted `previous_response_not_found` with the right code, at the right nesting depth, inside `error`. The compliance suite could read the code. The frame nonetheless failed schema validation on the envelope key name, and a failed frame validation fails the whole turn.

## Spec

> WebSocket failures MUST be sent as a JSON `error` envelope with a `status` code and an `error.code`.

`2026-04-24.mdx:166` — the OpenAPI `WebSocketErrorEvent` schema requires `["type", "status", "error"]`.
https://github.com/openresponses/openresponses/blob/92c12d96d7b61d6d15e2214daa5e9c6000ab6e1c/src/specifications/2026-04-24.mdx#L166

## What changed

- `sendError` in `packages/gateway/src/data-plane/chat/responses/websocket.ts` emits `status` instead of `status_code`, and its parameter is renamed `statusCode` → `status` to match.
- The reference URL for the required key set now sits on `sendError` itself.
- Two doc comments described the frame as "close-frame-shaped". It is not a close frame — it is the spec's error envelope, sent over an open socket — so that wording is gone.
- `CHANGELOG.md` records the rename as `minor`: clients reading `error.code` need no adaptation; clients reading the numeric status off the envelope must read `status`.

## Verification

Observed on an integration branch containing the full OpenResponses fix set, **not** on this branch alone: the 17-test OpenResponses compliance suite passed 17/17 on all three Responses paths (`gpt-5-mini` native, `gpt-4.1` via Chat Completions, `claude-haiku-4-5` via Messages), twice, with typecheck and lint clean and 4935 unit tests passing. Baseline before the work was 1/17 per path.

For this WebSocket stack specifically, the six WebSocket tests went 0/6 → 1/6 per path, with `response.done`, `ping` and `error` frame-validation failures reaching zero and the eviction assertion passing. What still fails on those tests is the envelope gap alone, addressed elsewhere in the set.

## Classification

**Category 1 (clean).** A private key name is renamed to the one the spec requires. There is no competing prior decision — `status_code` was never justified in writing, had no second producer, and the error payload it wrapped was already correct.

## Note on this diff

This branch carries merges of `main` from a window in which `main` also held the envelope-completion, protocol-type-truth, and compaction-resource work that was subsequently reverted off `main`. Until those land again, this PR's diff against `main` shows their files too (`envelope.ts`, `client-envelope.ts`, `to-result.ts`, and their tests). The change owned by this PR is the `websocket.ts` envelope key and its `CHANGELOG.md` entry; the diff narrows on its own once the siblings are back on `main`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ Impact levels:
 
 Hard and minor entries may include recommended actions; those actions do not need a separate advisory entry.
 
+## 2026-07-30 · minor
+
+### Responses WebSocket error frames carry `status` instead of `status_code`
+
+The JSON error envelope sent on the Responses WebSocket transport renamed its HTTP-style status key from `status_code` to `status`, matching the OpenResponses 2026-04-24 `WebSocketErrorEvent` contract. The nested `error` object (`type`, `code`, `message`, `param`) is unchanged, so clients that only read `error.code` need no adaptation; clients that read the numeric status off the envelope must read `status`.
+
 ## 2026-07-24 · advisory
 
 ### Audit dump files created before payload-file tracking
--- a/packages/gateway/__tests__/data-plane/chat/responses/envelope_test.ts
+++ b/packages/gateway/__tests__/data-plane/chat/responses/envelope_test.ts
@@ -1,0 +1,301 @@
+import { describe, it } from 'vitest';
+
+import { completeResponsesEnvelope, wrapResponsesEnvelopeCompletion } from '../../../../src/data-plane/chat/responses/envelope.ts';
+import { doneFrame, eventFrame, type ProtocolFrame } from '@floway-dev/protocols/common';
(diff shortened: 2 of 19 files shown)


Menci:
Superseded by #319, which points at the same commit (`94c1ee04c`) through the merge-free ref `pr/responses-ws-error-status`. This PR's head carried a `Merge branch 'main'` from a window when main still held the reverted sibling squashes, so its diff showed their files too. No code differs.

PR #307, opened 2026-07-30
fix(responses): make the terminal event the last WebSocket frame of a turn

Menci:
Second of four PRs recreating the WebSocket conformance stack against the OpenResponses `2026-04-24` specification. Stacked on #306 (`responses-ws-error-status`); draft until its base is `main`.

## Defect

The WebSocket transport flushed the terminal response event (`response.completed` / `response.failed` / `response.incomplete`) the moment it arrived, drained the rest of the generator, and then appended a Floway-private `{ "type": "response.done", "response": { id, usage } }` envelope.

Two things are wrong with that. `response.done` is not a member of the spec's streaming-event union, so every turn ended on a frame no compliant client can validate. Worse, because it was sent **after** the terminal event, the compliance runner — which reads frames in order and treats the terminal event as the turn boundary — attributed the trailing frame to the **next** turn. That failed every multi-turn WebSocket test, including ones whose own assertions were satisfied.

## Spec

- `2026-04-24.mdx:84` — the `[DONE]` sentinel is SSE-only and has no WebSocket counterpart.
- `2026-04-24.mdx:117` — WebSocket carries the same streaming event objects as streaming HTTP. `response.done` is not one of them, and the spec defines no frame after the terminal event.

https://github.com/openresponses/openresponses/blob/92c12d96d7b61d6d15e2214daa5e9c6000ab6e1c/src/specifications/2026-04-24.mdx#L117

## What changed

- The terminal event is buffered rather than sent inline, and flushed once the loop has run to completion — so it is the last frame of the turn.
- `response.done` and its `responseDoneSummary` helper are deleted.
- `completion` bookkeeping simplifies accordingly: the terminal flush is the single place that can downgrade a turn to `cancel`.
- Tests in `websocket_test.ts` and `codex/routes_websocket_test.ts` assert the terminal event as the final frame instead of `response.done`.
- `CHANGELOG.md` records it as `minor`: clients that waited for `response.done` before sending the next `response.create` must wait for the terminal event; its `response.id` is the id to continue from.

## Why the guarantee survives

`response.done`'s code comment advertised a real property: it was the stable signal that the turn's item and snapshot writes had committed, so a follow-up message could reference the stored response. That property is not lost. The buffered terminal event is emitted only after the wrapped writes have committed **and** the generator is exhausted, so its arrival carries exactly the same guarantee. And a client cannot race it: the `onMessage` queue chain serialises turns on the socket, so a follow-up `response.create` cannot be processed before the previous turn's writes settle regardless of what the client does with the frame.

## Verification

Observed on an integration branch containing the full OpenResponses fix set, **not** on this branch alone: 17/17 on all three Responses paths (`gpt-5-mini` native, `gpt-4.1` via Chat Completions, `claude-haiku-4-5` via Messages), twice, typecheck and lint clean, 4935 unit tests passing. Baseline was 1/17 per path.

For this stack: the six WebSocket tests went 0/6 → 1/6 per path; `response.done`, `ping` and `error` frame-validation failures reached zero.

## Classification

**Category 2 (contradicts an explicit prior decision but works).**

`response.done` was not an accident. It was documented in `websocket.ts` as a deliberate, stable signal for referencing a stored response, and that intent is recorded in code. This PR overturns that decision.

The grounds: the frame is outside the spec's event union, it is positionally indistinguishable from the next turn's first frame for an ordered reader, and the guarantee it was created to provide is already structural via the message queue. Nothing is given up. But a reviewer should register that a written decision is being reversed rather than a bug being fixed, and the reversal deserves an explicit yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ Hard and minor entries may include recommended actions; those actions do not nee
 
 ## 2026-07-30 · minor
 
+### Responses WebSocket turns end on the terminal event, and `response.done` is gone
+
+The Responses WebSocket transport no longer sends a trailing `{ "type": "response.done" }` frame. That frame was a Floway extension outside the OpenResponses streaming-event union; the terminal event (`response.completed`, `response.failed`, or `response.incomplete`) now carries the guarantee it advertised — it is flushed only after the turn's item and snapshot writes have committed and the event stream has drained, and it is the last frame of the turn. Clients that waited for `response.done` before sending the next `response.create` must wait for the terminal event instead; its `response.id` is the id to continue from.
+
+## 2026-07-30 · minor
+
 ### Responses WebSocket error frames carry `status` instead of `status_code`
 
 The JSON error envelope sent on the Responses WebSocket transport renamed its HTTP-style status key from `status_code` to `status`, matching the OpenResponses 2026-04-24 `WebSocketErrorEvent` contract. The nested `error` object (`type`, `code`, `message`, `param`) is unchanged, so clients that only read `error.code` need no adaptation; clients that read the numeric status off the envelope must read `status`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ Impact levels:
 
 Hard and minor entries may include recommended actions; those actions do not need a separate advisory entry.
 
+## 2026-07-30 · minor
+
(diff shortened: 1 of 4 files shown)


Menci:
Superseded by #320 (merge-free head `pr/responses-ws-terminal-last`, commit `7a11082ae`). No code differs.

PR #308, opened 2026-07-30
fix(responses): complete the client-facing response envelope at the egress boundary

Menci:
> Stacked on #305 (`responses-protocol-type-truth`). Draft until its base is on `main`.

## The defect

`POST /responses` did not return a spec-conformant response resource on any path. On the two translated paths the body carried only `id, object, model, output, output_text, status, error, incomplete_details, usage` — `responseResourceSchema` requires **31 keys**, so twenty-two were missing, plus `usage.output_tokens_details`. The native Copilot path was otherwise complete but never carried `prompt_cache_key`. The same gap appeared inside **every streaming `response.*` envelope event**, not just the non-streaming body.

## The seam

All three paths already converge on `wrapNativeResponsesClientOutput` in `client-output.ts`, called from the HTTP, WebSocket and compact entry points. Completion belongs there, at the last gateway stage before serialization.

Pushing it into the translators would be wrong on both practical and architectural grounds. Practically, `translate` knows neither `created_at` (which must be the gateway's request-start instant, not `Date.now()` inside a codec), nor the generated response id, nor the store decision, nor the account's service tier — and the table would be duplicated in `responses-via-chat-completions` and `responses-via-messages` while the native path still needed its own copy. Architecturally, a translator's output is *interior*: it feeds the server-tool shim, affinity egress and the item store, and it re-enters `responses.result()` on a stitched multi-call turn.

## One semantic, selected by one schema bit

The merge has a single semantic, chosen per key by whether the required slot admits `null`:

- `stated(candidates, fallback)` for the **20 non-nullable** required keys — the schema gives absence no spelling, so a stated default terminates the chain.
- `observed(candidates)` for the **11 nullable** ones — `null` *is* the schema's spelling for absence, so only presence is needed.

Candidate order follows the resource's own authority: what the upstream reported outranks what the client asked for, because every description on these keys is past tense ("the tools that **were available** to the model during response generation", "the service tier that **was used**").

Total normalizers then run on the **resolved** value. That last detail is not cosmetic. It fixes a real latent conformance bug: `completeTool` previously ran only on the echoed branch while `...response` spread *after* it, so an upstream-echoed or shim-restored (`restoreEchoedTools`) tool array reached the client missing `description`/`parameters`/`strict`. The native path passed only because Copilot happens to echo `strict: true`. A regression test pins it.

## Types are derived, not hand-written

`ClientResponsesEnvelope` / `ClientResponsesStreamEvent` are **derived** from `ResponsesResult` / `ResponsesStreamEvent` via a mapped type with `-?` plus a distributive conditional — never hand-written twins, so they cannot drift, and the ~34 other files naming `ResponsesResult` are untouched (the narrow types are assignable to the wide ones and generators are covariant in their yield type).

Exactly three signature changes plus a `sendResponsesEvent`/`sendJson` split:

1. `wrapResponsesEnvelopeCompletion` yields `ProtocolFrame<ClientResponsesStreamEvent>`.
2. `wrapNativeResponsesClientOutput` returns the narrow frame type.
3. `collectResponsesProtocolEventsToResult` gains a **narrow-first overload**. Implementation note worth recording: this needs two overloads rather than a type parameter, because TypeScript cannot infer a type argument out of a distributive conditional — a generic `R` would silently fall back to its default.
4. The WebSocket turn's frames go out through `sendResponsesEvent`, which accepts only a completed event; the transport's own `error` and keep-alive frames stay on the untyped `sendJson`.

### Files

`packages/protocols/src/responses/client-envelope.ts` (new), `.../responses/index.ts`, `.../responses/to-result.ts`; `packages/gateway/src/data-plane/chat/responses/{envelope.ts (new),client-output.ts,http.ts,respond.ts,serve.ts,websocket.ts}`, with `envelope_test.ts`, `http_test.ts` and `websocket_test.ts`.

## The decision conflict — stated honestly

Commit `44322150d` (#125, 2026-06-29) says verbatim:

> the spec defaults observed via Azure echo (all 'auto') signal 'backend decides' rather than canonical values worth synthesizing

and

> absent-echo (no tools synthesized when upstream omits it)

It built `upstreamResponseSnapshot` / `overlayOnSnapshot`, which **throw rather than fabricate**. This PR synthesizes exactly those values.

The tension is unavoidable rather than a matter of taste: `responseResourceSchema` types the contested keys **required and non-nullable**, so a conformant envelope can neither omit them nor send `null`. The choice is between a stated default and a body that fails validation. #125 was never asked that question, because it was not producing a client-facing body.

The reconciling boundary — stated in a comment at the top of `envelope.ts` — is:

> **Interior stages never state a value they did not observe. The egress stage is the only place Floway states one, and only where the schema forbids both omission and `null`.**

What makes that safe is that `commitSnapshot` persists **ordered item ids only** and runs *before* completion, so no invented value is ever persisted; telemetry reads `billableUsage`, not the envelope; affinity egress runs earlier. Nothing downstream of completion is Floway.

**Say it plainly:** for the invented-default half, this **overturns** #125's absent-echo rule rather than extending it. The boundary explains where the two rules apply and why both are defensible, but on the 20 non-nullable required keys the behaviour is the opposite of what #125 decided, and it is decided that way deliberately because the wire schema leaves no third option.

There is also an older, related finding worth citing: a **2026-05-26** pre-checkin review rated this same envelope sparseness **Critical** — *"only `id`/`model`/`object`/`status`/`output`/`output_text`/`usage`/`error` survive"*. Only the `error`/`incomplete_details` half was ever adopted; this PR is the rest of that answer.

## Classification — Category 2 (contradicts an explicit prior decision but works)

It overturns #125's absent-echo rule for the 20 non-nullable required keys, with the interior-vs-egress boundary as the reconciliation. Note also that it was **reworked** from an earlier shape a review had called architecturally unclean — three overlapping tables, three merge semantics fighting inside one object spread, and the echoed-tools bypass. The version here has one table-free semantic and no bypass.

## Verification

Observed on an integration branch containing the full OpenResponses fix set, not on this branch alone: 17/17 on all three Responses paths (`gpt-5-mini` native, `gpt-4.1` via Chat Completions, `claude-haiku-4-5` via Messages), twice, with typecheck and lint clean and 4935 unit tests passing. Baseline was 1/17 per path.

For this change specifically: **every envelope-level error disappeared on all three paths**, and the native path's filtered subset went **0/5 → 5/5**.

Supersedes #299, which was merged without authorization and cannot be reopened.

### DIFF
--- a/packages/gateway/__tests__/data-plane/chat/responses/envelope_test.ts
+++ b/packages/gateway/__tests__/data-plane/chat/responses/envelope_test.ts
@@ -1,0 +1,301 @@
+import { describe, it } from 'vitest';
+
+import { completeResponsesEnvelope, wrapResponsesEnvelopeCompletion } from '../../../../src/data-plane/chat/responses/envelope.ts';
+import { doneFrame, eventFrame, type ProtocolFrame } from '@floway-dev/protocols/common';
+import type { CanonicalResponsesPayload, ResponsesResult, ResponsesStreamEvent } from '@floway-dev/protocols/responses';
+import { assertEquals, assertExists } from '@floway-dev/test-utils';
+
+// Every key `ResponseResource.required` lists.
+// https://github.com/openresponses/openresponses/blob/92c12d96d7b61d6d15e2214daa5e9c6000ab6e1c/public/openapi/openapi.json#L2691-L2723
+const REQUIRED_ENVELOPE_KEYS = [
+  'id',
+  'object',
+  'created_at',
+  'completed_at',
+  'status',
+  'incomplete_details',
+  'model',
+  'previous_response_id',
+  'instructions',
+  'output',
+  'error',
+  'tools',
+  'tool_choice',
+  'truncation',
+  'parallel_tool_calls',
+  'text',
+  'top_p',
+  'presence_penalty',
+  'frequency_penalty',
+  'top_logprobs',
+  'temperature',
+  'reasoning',
+  'usage',
+  'max_output_tokens',
+  'max_tool_calls',
+  'store',
+  'background',
+  'service_tier',
+  'metadata',
+  'safety_identifier',
+  'prompt_cache_key',
+] as const;
+
+const request = (overrides: Partial<CanonicalResponsesPayload> = {}): CanonicalResponsesPayload => ({
+  model: 'gpt-5-mini',
+  input: [{ type: 'message', role: 'user', content: 'hi' }],
+  ...overrides,
+});
+
+const sources = (overrides: Partial<CanonicalResponsesPayload> = {}) => ({
+  request: request(overrides),
+  createdAt: 1_700_000_000,
+  stored: true,
+});
+
+// The shape a translated target hands to the client boundary: identity,
+// status, output and usage, and nothing else.
+const translatedEnvelope = (): ResponsesResult => ({
+  id: 'resp_1',
+  object: 'response',
+  model: 'gpt-4.1',
+  output: [],
+  status: 'completed',
+  error: null,
+  incomplete_details: null,
+  usage: { input_tokens: 8, output_tokens: 10, total_tokens: 18 },
+});
+
+const missingKeys = (envelope: Record<string, unknown>): string[] =>
+  REQUIRED_ENVELOPE_KEYS.filter(key => !(key in envelope));
+
+describe('Responses envelope completion', () => {
+  it('fills every required key a translated target left out', () => {
+    const completed = completeResponsesEnvelope(translatedEnvelope(), sources(), true);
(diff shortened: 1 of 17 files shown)


Menci:
Superseded by #323.

This change is unchanged in substance; the branch was rewritten to drop the word "envelope" in favour of the spec's own noun (`ResponseResource`), which renamed the module, the derived types and the wrapper. GitHub cannot repoint a PR's head branch, so the work moved to the new ref `pr/responses-resource-completion` and this PR is closed in favour of #323. Same base (`responses-protocol-type-truth`, #305), same draft status.

Posted by the cleanup agent that performed the rename.

PR #309, opened 2026-07-30
fix(responses): stop emitting JSON `ping` keep-alive frames

Menci:
Third of four PRs recreating the WebSocket conformance stack against the OpenResponses `2026-04-24` specification. Stacked on #307 (`responses-ws-terminal-last`); draft until its base is `main`.

**This is the PR in the stack with a real product tradeoff.** It buys wire conformance at the cost of idle-connection survival on Cloudflare, and the mechanism that would give us both is unreachable on that runtime. Read the tradeoff section before the diff.

## Defect

The WebSocket loop raced each pending upstream frame against a 15 s timer and sent `{"type":"ping"}` whenever the timer won. `ping` is not a member of the Responses streaming-event union, so any turn with a gap longer than the interval made the whole stream unvalidatable. Reproduced on `gpt-5-mini` at high reasoning effort: the first upstream frame lands at +32 s and a ping follows every 15 s after it.

`2026-04-24.mdx:117` — WebSocket carries the same streaming event objects as streaming HTTP; `ping` is not among them.
https://github.com/openresponses/openresponses/blob/92c12d96d7b61d6d15e2214daa5e9c6000ab6e1c/src/specifications/2026-04-24.mdx#L117

## What changed

- The race and `nextFrameOrKeepAlive` are deleted; the loop awaits the pending frame directly. The downstream abort controller, not a polling timer, is what unblocks the loop when the client goes away.
- `WsFrameRaceResult` becomes `WsFrameResult` (two variants, no `keep-alive`). `pendingWsFrameResult` stays, and now carries a comment explaining why it exists at all: the next frame is pulled before the current one is sent, so folding the rejection into a value attaches a handler synchronously and keeps a mid-stream upstream failure from surfacing as an unhandled rejection.
- The SSE transport is untouched. Its keep-alive is an SSE comment line — invisible to the event stream, and the only way to hold an idle HTTP response open.
- `CHANGELOG.md` records the removal and the client/operator actions below.

The branch's third commit (`db642f979`) deliberately narrows scope: a `ping` that `responses-via-messages` synthesized from an upstream Anthropic one was riding along here. That is source-protocol event-union hygiene, applies equally to SSE, and has nothing to do with WebSockets — it moved out to `responses-messages-ping-hygiene` (#301) so the two decisions are judged separately.

## The tradeoff

The keep-alive was not incidental and was not copied from SSE. It was added deliberately by #57 / `7345f2806` to fix a **real reported failure**: upstream SSE `ping` events are filtered out by `parseResponsesStream` (`packages/protocols/src/responses/stream.ts:76`), so during long reasoning the socket carries no frames at all, and Cloudflare's edge closes a WebSocket idle in both directions.

Measured, against a Cloudflare-proxied endpoint:

- torn down at TCP level at exactly **125.0 s**, with no CLOSE frame;
- **zero** edge-originated ping frames in that window;
- a second probe pinging every 30 s survived past 200 s.

So removing the JSON frame removes the only thing keeping a silent Cloudflare-hosted socket alive.

The correct mechanism — an RFC 6455 §5.5.2 ping/pong control frame — is **unreachable from a Worker**:

- `@cloudflare/workers-types`' `WebSocket` exposes only `accept` / `send` / `close` / `serializeAttachment` / `deserializeAttachment`;
- workerd matches: https://github.com/cloudflare/workerd/blob/26b5461b7dcc640bb16072f1ba6f2c6df82572ba/src/workerd/api/web-socket.h#L346-L394
- kj states the omission as a design decision — "Ping/Pong … are not exposed through this interface … The implementation is, however, expected to reply to Ping messages it receives": https://github.com/capnproto/capnproto/blob/e9fa5c7dc98192fc0dc0098ec770db68f997a938/c%2B%2B/src/kj/compat/http.h#L622-L631
- https://github.com/cloudflare/workerd/issues/3664 has been open since 2025-03.

Node **can** ping, but Node is not where the exposure is — nothing between the client and a self-hosted Node process enforces an idle timeout. The exposure is exactly the Cloudflare production target.

What this leaves as the interim story, per Cloudflare's own guidance (https://developers.cloudflare.com/network/websockets/): the keep-alive is delegated to the client. A client protocol ping roughly every 30 s resets the edge's idle timer because the runtime auto-replies with a pong. Enterprise accounts can additionally request a custom WebSocket idle timeout from Cloudflare support.

**Blast radius.** Codex CLI sends no heartbeat, and a protocol-level ping would not have helped it either, since the frame was never surfaced to it as a liveness signal. Any WS stream failure trips a session-wide fallback to HTTP SSE, so the failure mode is degradation, not breakage.

## Follow-up — not implemented in this PR

Reinstate the keep-alive behind an **explicit opt-in** — a subprotocol token or a query parameter — so a client that wants liveness over conformance can ask for it. That satisfies both the spec's event union (a client that did not opt in sees only union members) and Cloudflare-hosted sessions (a client that did opt in survives a silent turn). This is the proper resolution and it is deliberately out of scope here.

## Verification

Observed on an integration branch containing the full OpenResponses fix set, **not** on this branch alone: 17/17 on all three Responses paths (`gpt-5-mini` native, `gpt-4.1` via Chat Completions, `claude-haiku-4-5` via Messages), twice, typecheck and lint clean, 4935 unit tests passing. Baseline was 1/17 per path. For this stack, the six WebSocket tests went 0/6 → 1/6 per path; `response.done`, `ping` and `error` frame-validation failures reached zero.

**Stated honestly:** the keep-alive removal is timing-dependent, and no verification run hit a >15 s idle gap. The runs therefore neither confirm nor refute behaviour under slow reasoning — they are evidence that conformance improved, not evidence that idle survival is fine.

## Classification

**Category 4 (possibly should not be merged).**

This buys wire conformance at the cost of idle-connection survival on the runtime where Floway actually ships, and the correct mechanism on that runtime does not exist. It reverts a fix for a real, reported failure, and the compensating measure is an instruction to clients rather than something Floway does. The conformance argument is sound and the code is correct; whether the trade is worth taking before the opt-in follow-up exists is a judgement call for the reviewer, not for the author.


Menci:
Superseded by #321 (merge-free head `pr/responses-ws-no-json-ping`, commit `db642f979`). No code differs.

PR #310, opened 2026-07-30
fix(responses): evict a failed continuation's previous_response_id from connection-local state

Menci:
Fourth of four PRs recreating the WebSocket conformance stack against the OpenResponses `2026-04-24` specification. Stacked on #309 (`responses-ws-no-json-ping`); draft until its base is `main`.

## Defect

The spec makes eviction mandatory: a continuation turn that fails with a 4xx or 5xx error must drop the referenced `previous_response_id` from connection-local state, so a later attempt to continue from that id on the same socket fails with `previous_response_not_found`. Floway kept the snapshot, so a `store:false` chain silently survived a failed turn.

## Spec

> If a continuation turn fails with a `4xx` or `5xx` error, the server MUST evict the referenced `previous_response_id` from the connection-local cache. A later attempt to continue from that evicted `store=false` response ID on the same connection MUST fail with `previous_response_not_found`.

`2026-04-24.mdx:127`
https://github.com/openresponses/openresponses/blob/92c12d96d7b61d6d15e2214daa5e9c6000ab6e1c/src/specifications/2026-04-24.mdx#L127

The surrounding rule at `:125` — connection-local state is the only fallback when `store=false` — is what scopes the eviction.

## What changed

- `MemoryStatefulResponsesBacking` gains `evictSnapshot`, and `createResponsesWsSession` exposes it on the session handle. This is deliberately **beyond** the backing contract: the `StatefulResponsesBacking` interface and the durable backing gain no delete path, because the spec scopes eviction to connection-local state and persisted state stays addressable by id. Both spec quotes sit on the code in `items/store.ts`.
- `handleClientMessage` grows one turn-scoped `ResponsesWsTurnFailure` helper with `evict()` and `fail(status, error)`. The seven places a turn can fail — auth rejection, unsupported message type, `PreviousResponseNotFoundError`, `TranslatorInputError`, `WebSocketClientMessageError`, the catch-all 500, and the `api-error` / `internal-error` result branches — all route through `fail`, so the eviction cannot drift away from the error it belongs to. `sendError`'s `eventId` and `dump` arguments are now supplied by the helper rather than repeated at each call site.
- The turn's `previous_response_id` is captured *before* `expandPreviousResponseId` strips it off the payload, so a failing turn evicts exactly the id it referenced.
- The `finally` block calls `evict()` when the turn reports failure through a `response.failed` **terminal event** — that path answers the client with a streaming event, not an error envelope, so it never passes through `fail`.
- `websocket_test.ts` covers the eviction. `CHANGELOG.md` records it as `minor`: a chain that previously survived a failed turn now fails the retry with `previous_response_not_found`, and clients recover by starting a new response without `previous_response_id`. `store:true` chains with a non-zero retention window are unaffected.

## Known nit, stated up front

Eviction is triggered from two places: the failure helper and the `finally` covering a `response.failed` terminal. That is by construction — the two paths answer the client differently and neither subsumes the other — but it does mean the invariant is enforced twice, and a turn that fails through `fail` and then also settles as failed will call `evictSnapshot` twice. The operation is idempotent (`Map.delete` on an already-absent key), so the double trigger is a tidiness question, not a correctness one. Worth a reviewer's opinion on whether a single settle-time trigger would read better.

## Verification

Observed on an integration branch containing the full OpenResponses fix set, **not** on this branch alone: 17/17 on all three Responses paths (`gpt-5-mini` native, `gpt-4.1` via Chat Completions, `claude-haiku-4-5` via Messages), twice, typecheck and lint clean, 4935 unit tests passing. Baseline was 1/17 per path.

For this stack: the six WebSocket tests went 0/6 → 1/6 per path, `response.done` / `ping` / `error` frame-validation failures reached zero, and this PR's eviction assertion passes on all three paths.

## Classification

**Category 1 (clean)**, with the double-trigger nit above noted.

The change implements a spec MUST that Floway did not implement, adds no delete path where the spec does not scope one, and consolidates seven scattered failure exits into one helper on the way. No prior decision is overturned.


Menci:
Superseded by #322 (merge-free head `pr/responses-ws-evict-failed-continuation`, commit `71fe5e294`). No code differs.

PR #359, opened 2026-08-04
fix(translate): let a Chat client turn reasoning off on a Messages upstream

Menci:
## Problem

A Chat Completions client could not turn reasoning off on a Messages upstream. `reasoning_effort: 'none'` was filtered out of `output_config.effort` and then discarded, so the intent never reached the target.

Anthropic splits across two slots what OpenAI keeps on one. `output_config.effort` carries a level — `low | medium | high | xhigh | max`, with [no `none` member](https://platform.claude.com/docs/en/build-with-claude/effort) — and `thinking: {type: "disabled"}` is the off switch. Filtering `'none'` out of the effort slot was therefore correct; not putting it back on `thinking` was the bug.

The gap was one-sided. `responses-via-messages` already translated `'none'`, and `shared/messages-via/reasoning-effort.ts` already maps the other direction (`thinking.type === 'disabled'` → `'none'`), so the Chat pair also broke the round trip: a request carrying `thinking: {type: "disabled"}` could come back through Floway with thinking enabled.

It stayed invisible on Claude 4.5 and earlier, which do not reason unless asked, so dropping the off switch was a no-op there. It surfaces on a thinking-by-default target. Against Kimi K3 through Moonshot's Anthropic-compatible endpoint (`think_efforts.default_effort: "max"`), `'none'` is the only switch the model has, and a Chat client got a full `reasoning_text` on the very first turn.

## Change

Both `*-via-messages` pairs now read one `shared/via-messages/reasoning-effort.ts`, the target-locked mirror of the existing source-locked `shared/messages-via/reasoning-effort.ts`. The two directions are inverses of each other and can no longer drift apart; a round-trip test pins that.

Only `'none'` is translated, because it is the one value with a dedicated native slot. Every other value — including `xhigh`, `max`, and any vendor-private level — still reaches `output_config.effort` verbatim, so the upstream keeps ownership of which levels it accepts.

## Behaviour on thinking-always-on models

Whether a target honours the off switch is not uniform across Claude ([per-model table](https://platform.claude.com/docs/en/build-with-claude/thinking-troubleshooting#supported-models)): Fable 5, Mythos 5, and Mythos Preview reject `thinking: {type: "disabled"}` with a 400, and Opus 5 accepts it only at effort `high` or below. On those models a Chat client's `reasoning_effort: 'none'` now returns 400 where it previously succeeded while silently ignoring the request.

That is the intended outcome rather than a regression to paper over:

- It is not new. `responses-via-messages` has translated `'none'` this way all along, so a Responses client already receives that 400 today.
- It matches the native path. A client posting its own `thinking: {type: "disabled"}` to `/v1/messages` against Fable 5 gets the same 400 forwarded verbatim. A translated request and a native one now get the same answer from the same model.

Swallowing `disabled` for those models is model-family knowledge, which per `AGENTS.md` may not live in `packages/translate`. It belongs in `provider-claude-code` or beside the existing `disable-reasoning-on-forced-tool-choice` interceptor, and is out of scope here.

## Test Plan

- [x] `pnpm run test` — 474 files, 5118 tests passed
- [x] `pnpm run lint` — clean (run with the eslint cache cleared)
- [x] `pnpm run typecheck` — clean
- [x] Live against `kimi-k3` on `https://api.moonshot.ai/anthropic` through a Node-target gateway: Chat `reasoning_effort: 'none'` returns content with no `reasoning_text`; `'high'` still returns `reasoning_text`; the Responses pair still emits a summary-less reasoning item
- [ ] Live against a thinking-always-on Claude model (Fable 5 / Mythos 5) to observe the 400 — no credential for those models available here; the behaviour is taken from Anthropic's published per-model table linked above


Menci:
Agent note (Claude Code, opened this PR): correcting the scope claim above after checking Anthropic's primary docs.

**`reasoning_effort: 'none'` is not uniformly serviceable on Claude**, so this PR changes failure behaviour on some models rather than being a pure improvement everywhere.

Anthropic's effort scale is `low | medium | high | xhigh | max` with no `none` member ([effort](https://platform.claude.com/docs/en/build-with-claude/effort)), so filtering `'none'` out of `output_config.effort` was correct. The only off switch is `thinking: {type: "disabled"}`, and it is model-gated ([per-model table](https://platform.claude.com/docs/en/build-with-claude/thinking-troubleshooting#supported-models)):

| Model | thinking default | `{type:"disabled"}` |
|---|---|---|
| Claude 4.5 and earlier thinking models | off | accepted |
| Claude Opus 5 | on | accepted only at effort `high` or below; `xhigh`/`max` → 400 |
| Claude Fable 5 / Mythos 5 / Mythos Preview | always on | 400 |

So on always-on models this turns a silent no-op into a hard 400. Two things make that defensible rather than a regression to fix:

1. It is not new. `responses-via-messages` has translated `'none'` this way all along, so a Responses client already gets the 400 today. This PR removes the divergence rather than creating it.
2. It matches the native path. A client posting its own `thinking: {type: "disabled"}` to `/v1/messages` against Fable 5 gets the same 400 forwarded verbatim. After this PR a translated request and a native one get the same answer from the same model.

The alternative — swallowing `disabled` for thinking-always-on Claude models — is model-family knowledge, which per `AGENTS.md` may not live in `packages/translate`. It would belong in `provider-claude-code` or beside the existing `disable-reasoning-on-forced-tool-choice` interceptor, and is out of scope here.

b04fc9c records both citations on the shared helper so the slot choice and the emit-the-intent decision are traceable. Flagging the behaviour change explicitly because it is the kind of thing that should be a deliberate call, not a side effect.

PR #380, opened 2026-08-05
fix(gateway): generate unbiased API key tokens

Menci:
## Problem

API-key characters were selected with `randomByte % 62`, giving the first eight Base62 characters a 25% relative bias.

## Change

- Generate both 20-character segments with Nano ID's rejection-sampling `customAlphabet`.
- Preserve the exact public `sk-{20}T3BlbkFJ{20}` format.
- Pin Nano ID 5.1 for Floway's complete Node runtime contract.

## Test Plan

- [x] API-key targeted suite
- [x] Gateway typecheck
- [x] Changed-file lint and diff checks
- [x] GitHub Verify — lint, typecheck, test, installers, generated-assets, build


m13253 (changes requested):
> Preserve the exact public sk-{20}T3BlbkFJ{20} token format.

Push back due to Hyrum’s Law: Do not depend on OpenAI’s undocumented behavior. API keys should be opaque ASCII strings and applications should never assume an API Key format. The `sk-` prefix is only a guard for PII filters. Also, I’ve never seen any client that rejects a non-OpenAI-formatted key for an OpenAI-compatible endpoint.

Proposal: `sk-{48}`, nothing more, nothing less.

PR #431, opened 2026-08-07
fix(docker): start the Podman Quadlet units independently

m13253:
Remove unit dependency between `floway-server.container` and `floway-web.container` so each of them can be restarted independently.


m13253:
This change should be harmless. But I will check whether the restart works when Floway crashes the next time on my machine.
After I confirm the auto-restart works correctly, I will finalize the pull request.

m13253:
My Floway crashed the third time from OOM, and this new configuration restarted it immediately. I think it is ready to merge.

PR #443, opened 2026-08-08
feat(ollama): report the Cloud account behind an upstream

Menci:
## What this adds

Ollama Cloud meters a subscription through a rolling session window and a weekly window. Floway showed neither, so an operator had to open ollama.com to see where they stood. The upstream editor now names the account behind the key and reports both windows.

## Reading the account

Nothing rides on the inference responses to harvest — Ollama Cloud sends no rate-limit headers — so both readings are active probes behind the API key the data plane already holds:

- `GET /api/usage` returns the two windows as fractions of the plan's allowance, plus the trailing activity cost.
- `POST /api/me` names the account and the plan those fractions are of.

Each has its own state slot and its own interval, because they move on different clocks. Every inference call the provider serves arms both, debounced against the attempt time persisted on the upstream: **one minute** for the windows, **one day** for the account, since a subscription changes on a billing boundary at most. Token counting never reaches a model and arms neither. The card's refresh reads both unconditionally — a day is too long to wait after fixing a key.

Both writes go through the state repo's read-modify-CAS and resolve by attempt time, so a race cannot roll a slot back to an older reading.

## Scope is the operator's answer

`config.cloudUsage` states whether an upstream is a Cloud account whose usage the gateway reads. A base URL cannot settle that in either direction: the cloud may be reached through the operator's own domain, and a private daemon may sit behind one that looks like anything. The dashboard answers it for them — editing the base URL to the cloud endpoint turns the option on, editing it away turns it off, and the operator's own toggle ends the suggestion for good. `0078_ollama_cloud_usage.sql` carries stored records forward: an upstream pointed at ollama.com comes forward with the option on.

## What the card shows, and what it does not

The windows are percentages with no countdown: the endpoint reports neither reset timestamps nor window lengths, and the labels name the periods from Ollama's own pricing page. The per-model request counts the endpoint also returns are stored but not shown — Ollama meters an allowance by tokens, not by requests, so a count beside a percentage reads as an explanation of it and is not one.

Only the plan, name, and email are kept from the account body. The avatar is dropped rather than have the operator's browser fetch an image from ollama.com on every visit.

A failed probe keeps the last good reading, stamps the attempt so a failing upstream is retried on the same cadence rather than on every request, and records the error where the operator sees it. It never reaches the request that armed it.

## Test Plan

- [x] `pnpm run verify` — lint, typecheck, 5560 tests across 538 files, installer tests, repository checks, and the web build
- [x] Live account on ollama.com: a real inference call through the gateway arms the probes, and both slots persist their readings
- [x] Debounce measured on real traffic: six calls inside one minute produce no second usage probe; a call after the window refreshes it; the account slot holds across an hour and re-reads after a day
- [x] Operator refresh reads both endpoints and surfaces an upstream failure in its own words
- [x] Card renders in light and dark, and the option's suggestion follows base-URL edits without overriding a saved upstream or the operator's own toggle
- [x] Migration test covers cloud, trailing-slash, self-hosted, mirror, already-chosen, and non-Ollama rows


Menci:
## What the two endpoints send

`GET /api/usage`, live:

```json
{"activity": {"cost": "0.00000",
              "period": {"type": "last_4_weeks", "starting_at": "2026-07-13T00:00:00Z", "ending_at": "…"},
              "models": []},
 "limits": {"session": {"usage": 0.001, "models": [{"name": "gpt-oss:120b", "request_count": 6}]},
            "weekly":  {"usage": 0, "models": […]}}}
```

Unannounced: absent from docs.ollama.com, whose [`/api/usage` page](https://github.com/ollama/ollama/blob/f0078ae4766d0d570e196158f20dde309bd96124/docs/api/usage.mdx) documents per-response performance counters instead. A maintainer [pointed users at it](https://github.com/ollama/ollama/issues/12532#issuecomment-5117276581) on 2026-07-29 and an account holder [posted the body back](https://github.com/ollama/ollama/issues/12532#issuecomment-5117969589); the request for it is [still open](https://github.com/ollama/ollama/issues/12532). No reset timestamps, and no window lengths — [ollama.com/pricing](https://ollama.com/pricing) is where "session limits that reset every 5 hours" comes from, and where usage is stated to be metered by model and by input, cached-input, and output tokens rather than by request count.

Nothing equivalent rides on inference responses: the [documented error contract](https://github.com/ollama/ollama/blob/f0078ae4766d0d570e196158f20dde309bd96124/docs/api/errors.mdx) is a bare 429 with a JSON `error` string, no rate-limit headers, no `Retry-After`.

`POST /api/me` (405 to GET), live:

```console
$ curl -s -X POST -H "Authorization: Bearer $OLLAMA_API_KEY" https://ollama.com/api/me | jq 'keys'
["AvatarURL","Bio","CreatedAt","Email","FirstName","ID","LastName","Links","Name","Plan"]
```

Capitalized. The client's [`UserResponse`](https://github.com/ollama/ollama/blob/f0078ae4766d0d570e196158f20dde309bd96124/api/types.go#L939-L949) declares `json:"plan"`, but that is the decoding type: the hosted service marshals a wider, untagged Go type — `CreatedAt` and `Links` are not in `UserResponse` at all — so it emits Go field names, and a reader keyed on those tags finds nothing.

## Prior art

Every third-party tool predates the usage endpoint and scrapes `ollama.com/settings` with a browser session cookie: [ontech7/ollama-usage](https://github.com/ontech7/ollama-usage), [florian-croiset/ollama-usage](https://github.com/florian-croiset/ollama-usage), [jacklee-code/ollama-cloud-quota-monitor](https://github.com/jacklee-code/ollama-cloud-quota-monitor). [slkiser/opencode-quota](https://github.com/slkiser/opencode-quota/blob/main/src/lib/ollama-cloud.ts) uses the endpoint, and reads the per-model rows under a different naming than the live body uses.

PR #452, opened 2026-08-08
feat(ollama): read the Cloud account's plan

Menci:
Stacked on #450 (itself stacked on #443), and a sibling of #451 — both add the one plan name the routing list could not yet show. Draft until its bases land.

## Why

After #450 an upstream row leads with the plan it serves on. Ollama Cloud led with the provider's name, because `/api/usage` reports windows and a cost and says nothing about the account.

`POST https://ollama.com/api/me` answers with the account and its `plan`. Unlike the usage endpoint beside it, this one is in the official Go client — which is what names the field and its values, `free` / `pro` / `max` / `team`:

- [`api/client.go` `Whoami` → `UserResponse.Plan`](https://github.com/ollama/ollama/blob/f0078ae4766d0d570e196158f20dde309bd96124/api/client.go#L506)
- [`server/routes.go` — an empty plan defaults to `free`](https://github.com/ollama/ollama/blob/f0078ae4766d0d570e196158f20dde309bd96124/server/routes.go#L2207)

Everything else was ruled out: `/api/usage` carries no plan field, and `/api/user`, `/api/account`, `/api/whoami`, `/api/subscription`, `/api/billing`, `/api/plan`, `/api/limits`, `/api/tier`, `/api/quota` and `/api/profile` all 404.

## What changes

- `ollamaFetchMe` beside `ollamaFetchUsage`; POST, because the route is registered for POST only and answers 405 otherwise.
- A second state slot, `account: { fetchedAt, plan }`, on its own refresh cadence.
- Armed by the same data-plane calls that arm the usage probe, behind `OLLAMA_ACCOUNT_PROBE_MIN_INTERVAL_MS` = 24h: a subscription changes on a billing boundary at most, so a busy upstream re-reads its plan once a day rather than once a minute.
- `apps/web/src/components/upstreams/ollama-account.ts` maps the plan onto `Ollama Free / Pro / Max / Team`, forwarding an unknown value as itself — these are plain words, not internal SKUs.

## Two deliberate asymmetries with the usage slot

- **No error field.** A usage window the probe could not read is something an operator acts on; a plan it could not read costs the row a name and nothing else. The failure is logged and the last known plan stands.
- **An absent `plan` is recorded as null, not as `free`.** The defaulting to free is the *server's*, in its own handler — an absent field on the wire is a fact about the response, not about the plan. The row then falls back to `Ollama`.

Only the one field is taken: the live body also carries subscription-period and billing identifiers that `UserResponse` does not declare, in a second casing.

## Verification

`lint`, `typecheck`, and the full suite pass (532 files / 5528 tests). New tests: the probe posts to the right URL with the upstream key and keeps the plan through a body richer than the declared type; an account naming no plan records none; a 401 leaves the last known plan in place; and, driven through the provider rather than the scheduler, one inference call reads both probes cold, neither when both are fresh, and only the plan when it is a day old beside minute-old windows.

### DIFF
--- a/apps/web/__tests__/components/upstreams/signals_test.ts
+++ b/apps/web/__tests__/components/upstreams/signals_test.ts
@@ -176,10 +176,24 @@ describe('upstream readout by provider', () => {
     expect(planFor('prolite')).toBe('ChatGPT Pro Lite');
   });
 
+  it('names the Ollama Cloud account by the plan the probe read', () => {
+    const planFor = (plan: string | null) => readoutOf({
+      kind: 'ollama',
+      state: { account: { fetchedAt: Date.parse(OBSERVED), plan } },
+    }).plan;
+
+    expect(planFor('max')).toBe('Ollama Max');
+    // The identifiers are plain words, so one this table has not seen reads as
+    // itself rather than as nothing.
+    expect(planFor('enterprise')).toBe('Ollama enterprise');
+    expect(planFor(null)).toBe('Ollama');
+  });
+
   it('reads the Ollama Cloud windows and the activity cost the probe stored', () => {
     const record = {
       kind: 'ollama',
       state: {
+        account: { fetchedAt: Date.parse(OBSERVED), plan: 'pro' },
         usageProbe: {
           attemptedAt: Date.parse(OBSERVED), error: null, observation: {
             fetchedAt: Date.parse(OBSERVED),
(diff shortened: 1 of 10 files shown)


Menci:
Superseded by #443, which now carries the account probe where it belongs — one PR with the usage windows it names the plan for.

Its version is also the correct one: it verified against a live account that ollama.com marshals an untagged Go type, so the body's keys are capitalized (`Plan`, not `plan`) and carry `CreatedAt`/`Links` that the client's own `UserResponse` does not declare. Reading the json tag names off that client type — which is what this PR did — gives a key the endpoint never sends.

The list-side plan label moves to #450.

PR #453, opened 2026-08-08
fix(web): name an upstream by its account rather than by its plan

Menci:
The row's two lines had drifted into saying the same thing twice and, on Ollama, into saying nothing at all.

- **Codex** led with `me@example.com - pro`, but the line beneath now opens with that plan for every provider that has one, so the identity line repeated it. It states the account alone.
- **Ollama Cloud** led with `https://ollama.com`, which is the same endpoint for every cloud upstream and so distinguished none of them. It now leads with the account the key belongs to — the thing an operator would tell two of them apart by. A self-hosted daemon has no account, so its address remains its whole identity.

| Upstream | Before | After |
|---|---|---|
| Codex | `me@example.com - pro` | `me@example.com` |
| Ollama Cloud | `https://ollama.com` | `me@example.com` |
| Ollama (self-hosted) | `http://127.0.0.1:11434` | unchanged |

An account with no email falls back to the same shortened identifier its own editor card shows, rather than the row's own separate truncation — the two surfaces name one account, so `shortAccountId` moves beside the badge and both read it.

## Test Plan

- [x] `pnpm run lint`
- [x] `pnpm run typecheck`
- [x] `pnpm run test` — 539 files, 5576 tests
- [x] `pnpm run build:web`
- [x] Read in a headless Chromium against a production build, with fabricated upstreams of every provider kind: the Codex and Ollama Cloud rows name their account, the self-hosted Ollama row keeps its address


Menci:
Folded into the single upstreams-page Pull Request, together with the rest of that page's work.

PR #483, opened 2026-08-18
feat(gateway): prefix non-Anthropic ids for the Claude Code picker

yyyr-p:
## Summary

The Claude Code CLI's gateway-discovery model picker filters `/v1/models` entries by id: Anthropic documents that it ignores any entry whose `id` doesn't begin with `claude` or `anthropic`, and the compiled `Bootstrap Gateway /v1/models` handler in `@anthropic-ai/claude-code@2.1.211` matches that verbatim — `models.filter(m => /^(claude|anthropic)/i.test(m.id))`, then a second pass that drops ids resolving to a built-in Claude family other than Fable. Any non-Anthropic model Floway advertises through discovery is silently dropped from the `/model` menu, so an operator who routes GPT-5 or Gemini through a gateway upstream never sees it selectable.

The `/v1/models` handler now prepends a `claude-code!` marker to every non-Anthropic id when the caller is the CLI (`claude-code/<version>` User-Agent). The picker admits the prefixed id, and because it renders `display_name ?? id`, the operator-configured label is still what the user sees — the id rewrite is invisible in the UI. The marker is an encoding prefix rather than a reserved namespace: a raw id that already begins with `claude-code!` gains another layer, so `M`, `P+M`, and `P+P+M` map to `P+M`, `P+P+M`, `P+P+P+M` and every advertised id stays unique without reserving `!` from upstream ids. The catalog is also narrowed to chat models before the rewrite, matching the Codex and Gemini discovery handlers — the picker is a chat surface and embedding/image entries only clutter it.

On the Messages path the same id comes back from the CLI's inference User-Agent (`claude-cli/<version>`, distinct from the `claude-code/<version>` discovery identity), and the serve boundary strips exactly one prefix layer before model resolution, so the generic candidate resolver stays unaware of the compatibility projection. Other clients' ids pass through untouched: the decode is gated on the `claude-cli/` User-Agent so a literal upstream id that happens to begin with `claude-code!` is preserved rather than truncated.

## Test Plan

- [x] `pnpm run verify` — the gateway suite is green, including 36 tests across `http_test.ts` and `serve_test.ts` covering the encode/decode round-trip, prefix doubling, the `[1m]` suffix composition, the chat-only narrow, and the non-Claude-Code passthrough. Five `apps/web` failures (Copilot device-flow polling, OAuth authorize-url) are pre-existing on `upstream/main` and unrelated to this PR — `apps/web` is unchanged in this branch.
- [x] A non-Anthropic id (`gpt-4o`) is served as `claude-code!gpt-4o` to the CLI and decoded back to `gpt-4o` before resolution; a literal `claude-code!gpt-4o` upstream id round-trips through a doubled prefix to its original form.


yyyr-p:
Wait for pipeline refactoring before merging this

yyyr-p:
Added support for Claude Desktop. This also solves #496

PR #491, opened 2026-08-19
feat: add custom OAuth2 Provider Login

pStrikeZ:
Only tested Gitea as provider


pStrikeZ:
Verbally rejected by the project maintainer

PR #492, opened 2026-08-19
feat: add adjustable site branding

pStrikeZ:


pStrikeZ:
Verbally rejected by the project maintainer

PR #514, opened 2026-09-17
fix(custom): carry chat metadata from auto-fetched models

m13253:
The custom provider parses the chat block (modalities, reasoning) off a Floway-shaped /models response, then projects each auto-fetched row into a ProviderModel without copying it. Manual rows already forwarded it, so an aggregating gateway serving a custom upstream advertised the wrong capabilities: limits, pricing and kind surfaced, but chat modality and reasoning metadata were dropped.

Carry model.chat through the auto projection and bump the persisted catalog revision so existing cached rows go cold and re-fetch.


m13253:
Known issues (will fix in other PRs):
* Automatically advertised metadata of model aliases are undetermined: Sometimes first candidate, sometimes last, sometimes empty. Needs triage and reprod.
  Workaround: Always set manual metadata for model aliases.
* Backing up and restoring aliases don’t work.
  Workaround: Manually copy and paste aliases across instances.

Menci (approved):

PR #516, opened 2026-09-18
fix(translate): surface vendor-dialect reasoning text in chat-completions output

m13253:
OpenAI Chat Completions has no reasoning-text field in its own spec; upstreams expose the same quantity as reasoning_content or reasoning. The chat-completions-source translators only read reasoning_text, so a stream carrying either alias dropped its reasoning when translated into OpenAI Responses, Anthropic Messages, or Gemini generateContent output.

Accept both aliases wherever reasoning_text is read, preferring the canonical name when an upstream emits more than one. The reassembler's field-fidelity contract is unchanged: an unknown field on a non-streaming chat-completions passthrough still reaches the client verbatim.


m13253:
Known issues: (will fix in separate PRs)
* Floway doesn’t support OpenResponses `reasoning_text`, which may cause roundtrip issues with certain client / server combinations, resulting in IQ degradation or cache rate reduction.
  **Evidence:** https://github.com/Menci/Floway/blob/1b15195adea3cef7798e5ec61a0c90904ed1704e/packages/protocols/src/openai-responses/index.ts#L206-L216
  **Workaround:** Please monitor cache hit rate. No fix yet.

